PR #3, opened 2016-09-22
Issue #2

petermlm:
I've implemented a few rules in the Makefile to handle the lib directory.

Note that the rules will also remove the lib directory  but will only due so if it is empty (in order not to delete any user files that might be there).


IanLee1521:
Looks good. Thanks @petermlm !

PR #5, opened 2016-12-15
Add support for CMake build files

chuckatkins:
I know I didn't discuss this with you first but I needed to implement a CMake build to integrate it with another application and it seemed suitable to upstream it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 *.a
 *.o
-examples/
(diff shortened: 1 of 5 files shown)


chuckatkins:
Closing to incorporate some of the differences between mine and @estan

estan:
Appreciate you taking the time to pick from my branch, glad I got something right at least :) I have a total lack of bragging right instincts, so pick and choose as you please, no need to credit me or have me submit them to your branch as PRs.

One thing I did which I don't know if Peter would approve of is to install the examples/zfp binary into bin/. I found it useful, and many compressors install a command line utility. But @lindstro, what do you say, should the `zfp` command in its current form be considered an example, or is it ready for prime time to be installed as the official zfp command line interface?

PR #7, opened 2017-03-14
cmake: Generate windows import library

chuckatkins:
I think this will fix the windows import library problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,9 @@ endif()
 # Top level options
 #------------------------------------------------------------------------------#
 
+# Use this to get a usable export library when building a DLL on Windows
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+
 # Suggest C99
 # Falls back to C89 if 99 is not supported
 set(CMAKE_C_STANDARD 99)


lindstro:
Thanks, Chuck.  I fixed a couple of missing extern "C" declarations in examples/fields.h, but have spent most of the day attempting to track down one more missing symbol:

C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\link.exe /ERRORREPORT:QUEUE /OUT:"C:\Users\lindstrom2\Documents\src\zfp\build\bin\Debug\testzfp-small.exe" /INCREMENTAL /NOLOGO kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib ..\lib\Debug\zfp.lib /MANIFEST /MANIFESTUAC:"level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG /PDB:"C:/Users/lindstrom2/Documents/src/zfp/build/bin/Debug/testzfp-small.pdb" /SUBSYSTEM:CONSOLE /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB:"C:/Users/lindstrom2/Documents/src/zfp/build/lib/Debug/testzfp-small.lib" /MACHINE:X86 /SAFESEH   /machine:X86 "testzfp-small.dir\Debug\testzfp.obj"
  "testzfp-small.dir\Debug\fields.obj"
testzfp.obj : error LNK2001: unresolved external symbol _stream_word_bits [C:\Users\lindstrom2\Documents\src\zfp\build\tests\testzfp-small.vcxproj]

This is a C const symbol defined in src/bitstream.c that should have external linkage and is declared as

  extern "C" const size_t stream_word_bits;

in bitstream.h (when this header is included in testzfp.cpp).  The symbol does appear in the dll but not in zfp.lib.  Any ideas?

lindstro:
I was able to determine that the reason for this missing symbol is that CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS exports only functions, not global data like the stream_word_bits constant (see https://cmake.org/cmake/help/v3.4/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html).  So I'm back to the drawing board.  I was able to get this to work with judicious insertions of __declspec macros and a smattering of ifdefs, but this gets messy when trying to handle both imported and exported symbols while trying to support both C and C++ and multiple compilers on top of zfp's unique use of C templates.  I will need to rethink my strategy here.

chuckatkins:
Ok, so scrap this then, it was worth a shot as an easy fix.  An easier way
to do this than the declspec stuff would be to create a .def file.  It's
basically just a text file with the list of symbols to export.  It's​ messy
to do for c++ but for c it's super easy.  You then just add the def file as
you would any other source file.  CMake will recognize it and pass the
right options to the compiler to use it.

On Mar 17, 2017 20:37, "Peter Lindstrom" <notifications@github.com> wrote:

I was able to determine that the reason for this missing symbol is that
CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS exports only functions, not global data
like the stream_word_bits constant (see https://cmake.org/cmake/help/
v3.4/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html). So I'm back to the drawing
board. I was able to get this to work with judicious insertions of
__declspec macros and a smattering of ifdefs, but this gets messy when
trying to handle both imported and exported symbols while trying to support
both C and C++ and multiple compilers on top of zfp's unique use of C
templates. I will need to rethink my strategy here.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/LLNL/zfp/pull/7#issuecomment-287502977>, or mute the
thread

lindstro:
I'm sure this is easy enough to set up, but it requires maintaining the def file as the zfp API evolves.  Instead, I've decided to do away with global constants and have them returned by functions.  zfp now compiles fine on Windows.

lindstro:
Never mind--I reverted back to using constants and appropriate __declspec declarations hidden behind platform generic macros.  In combination with CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS, this solves the DLL problem on Windows.  These changes are reflected in the latest zfp release (0.5.1), which enables support for Visual Studio builds.  I'm going to go ahead and close this PR.

chuckatkins:
@lindstro it looks like somebody fixed the "data symbols don't get exported" issue in upstream CMake: https://gitlab.kitware.com/cmake/cmake/merge_requests/763/commits .  It should be available in 3.9, if that helps at all.

lindstro:
@chuckatkins: Thanks for the pointer.  I'm sure we'll take advantage of this in the future.

PR #14, opened 2018-04-13
Misc cmake updates

chuckatkins:


lindstro:
Thanks for your contributions, Chuck.  We'll take a look at your PR.  I'll be on travel this coming week, so I likely won't get to it until some time after I return.

PR #15, opened 2018-05-01
Fix build warnigns

chuckatkins:
Resolve a variety of compiler warnings to allow building with gcc using ```-Werror```


chuckatkins:
Just for some context, this and the previous PR are in preparation for including zfp in ornladios/adios2 .  We've just turned on Werror in our CI so the third party libs we want to ship with (when not using an external version) need to pass it as well.

salasoom:
What compiler/version shows you the signed/unsigned mismatch warnings?

Those warnings come from tests which do not exist in our official releases. Is there a reason you are not using an official release with adios2? We aim for 2 releases per year, so these PR changes won't appear in an official release until September.

salasoom:
I figured out to use "-Wall -Werror" to reproduce the warnings, and fixed the rest of them. Will close and merge.

PR #20, opened 2018-10-11
Fix CUDA include flags

mohamed:
When compiling using CUDA, the inclusion to zfp.h was missing.
This patch fixes that by using cuda_include_directories()


mclarsen:
@mohamed Can you resubmit your pull request with develop as the base branch?

mohamed:
@mclarsen I changed the base. It should be now develop

PR #27, opened 2018-11-16
Fix linkage of endtoend tests under Linux

mohamed:
Under Linux, the linkage of the endtoend test was broken due to missing '-lrt'.
In addition, I found two typos in zfp_add_test which I fix in this commit


salasoom:
Thanks for catching these typos. Can you move the "find_library()" (and preceding comment) below the function, before it is repeatedly called?

Don't worry about the failing Appveyor build.

lindstro:
@salasoom, I'd like to know why we need to use clock_gettime, which is POSIX only, in one of our tests.  Certainly nothing should depend on the availability of a high-resolution timer.  Can we accomplish the same with a standard function, e.g., clock()?  See examples/speed.c, for instance.

salasoom:
@lindstro Those tests are not part of official releases, so they were only intended to run on CI machines. I can substitute clock() for wider compatibility.

@mohamed Disregard my first comment. I will remove the RT dependency.

lindstro:
Good point.  Still, they pose a portability problem for developers.  I agree that we should remove the rt dependency.

mohamed:
@salasoom I updated the commit. Till you remove the clock_gettime dependency, this commit should get develop back to green on Linux boxes with older glibc

salasoom:
I will amend this PR to remove all libRT-related code because we ultimately decided to proceed without it.

mohamed:
LGTM

PR #29, opened 2018-12-13
Python Bindings: Numpy Support

SteVwonder:
@salasoom and @lindstro: just wanted to post this PR to update you on the progress of the python bindings.  Support for compression/decompression from/to numpy arrays has been added, as well as a few tests to verify that a "round-trip" of compression/decompression works.  @salasoom, if you have some time before the holidays, it would be great to discuss what other tests would be beneficial.

TODO:

- [x] wrap main zfp functions and testing utility functions
- [x] add tests and add to Travis CI
- [x] floating-point (de-)compression for 1->4D matches known-good checksums
- [x] `write_header` argument for `compress_numpy`
- [x] add support for arbitrary strides rather than rearranging the array into fortran-contiguous
- [x] integer (de-)compression matches the known-good checksums
- [x] wrap the random strided array functions in test/utils and add to python tests
- [x] support decompression into a pre-allocated buffer
- [x] roll back changes to travis.sh
- [x] release GIL before calls to compress/decompress
- [x] update Cython flags to embed function signatures and docstrings
- [x] support for lossless compression
- [x] documentation
  - [x] dependencies, configuration, compilation, testing, and installation
  - [x] API docs and example usage
- [x] get tests passing on Travis
  - [x] Linux
  - [x] Mac
- [ ] breakup the large unittest method into finer granularity chunks
- [x] requirements.txt


SteVwonder:
Oh. It's probably worth mentioning that the bindings aren't built by default.  You have to provide `-DBUILD_PTYHON` on the cmake line to build them.

TODO: add a python test to travis

lindstro:
@SteVwonder This is fantastic!  I'm up against a paper deadline this week but either I and/or @salasoom will take a closer look over the next few days.

salasoom:
Can you base the PR from the `develop` branch? We exclude documentation and 99% of tests when making releases on `master`.

Now that we are finishing bindings in a couple languages, we may mess around with the directory structure in the next release. For now, please separate the tests into dir `tests/python`.

When on `develop`, you'll see all the tests (and docs). For testing array compression/decompression, we generate ~million element arrays and pass them into the compressor, and then checksum the resulting bitstream and decompressed array (once decompressed). You can find C libs in `tests/utils` for generating the smooth random arrays, and for checksum-ing resulting memory (hash32/64). Calling those from Python would seem to be the way to go. We compare checksums against what's stored in `tests/constants` (macro-defined constants).

I've been meaning to refactor some of those utility libraries, so your input can help re-shape those APIs to be the least painful when calling from Python (ex. removing double pointers).

As far as benchmarking goes, that's something we've been meaning to get around to, an executable we can run on local machines, because we see a large range of test runtimes on Travis and Appveyor. For now, we time how long compression and decompression take for those million element arrays, and print them out as part of the test. At least then we have a history of CI log output with some timings.

lindstro:
@SteVwonder I've had a chance to skim your PR.  I like a lot of what I'm seeing.  Great job!

I also have a few comments and concerns:

compress_numpy (which is perhaps the most important function) supports fixed-accuracy mode only, while decompress_numpy supports all modes.  Although we have to start somewhere, this is a significant limitation.  While we support only fixed-rate compression in the CUDA version, there are good technical reasons for this having to do with lack of knowledge where in the compressed stream each thread should start.  I know we'll be making changes to the Python API before the release, but I wanted to point out that we'll have to change this function to support the other modes.  For now, it might be best to rename the function to make it explicit that it only handles the fixed-accuracy mode.  On the other hand, only a single line (zfp_stream_set_accuracy) has to change to support fixed-precision and fixed-rate also.  Alternatively, we might consider passing in an object that encodes the compression mode, which is essentially what zfp_stream does, and then add three more short functions for setting the compression parameters.  I think Markus or I can make this change.

I don't think we should have a default tolerance.  The user should be aware that this is something they need to specify.  If they don't know how to set this, then I think it would be safer to use a tolerance of zero.

Another difference wrt. zfp_compress is that only one field can be compressed at a time.  That might be fine for the Python interface since one can simply make multiple calls and concatenate the compressed byte strings.  However, this does cause an incompatibility with files compressed via the C interface, as zfp_compress does not insert a header between each field.  This is something worth documenting at least.

Does Cython expose internal implementation functions like _init_field in the global namespace?  That would violate the ANSI standard, which reserves all public symbols starting with _.

class Memory(): Maybe check the return value of malloc?

The return value to zfp_decompress is not checked.  It will be zero in case of failure.  Maybe throw an exception?

One potential concern: By default, zfp's compressed data is in multiples of 64-bit "words."  Are there any concerns that Python might pass an unaligned array of bytes to decompress_numpy?  That could cause issues on some platforms.

Regarding the changes to the LICENSE, LLNL has gotten picky about what can go in this file: see https://software.llnl.gov/about/licenses/.  We may have to update the license and place this additional information in the python directory.

lindstro:
@SteVwonder Currently, there's an assumption that the numpy array uses C ordering, which may or may not be true.  I think it would be safer to call zfp_set_stride_* in _init_field by translating the ndarray.strides byte strides to zfp's scalar strides.  See this discussion regarding this issue: https://github.com/zarr-developers/numcodecs/pull/160.

SteVwonder:
@lindstro, I looked into the `type` and `exec` problems reported in zarr-developers/numcodecs#160.  I had no issues with referencing the `type` field in the `zfp_field` struct (it is already included in this PRs commits), but I did have an issue referencing the `exec` field in the `zfp_stream` struct (Cython would error out with "invalid syntax").

I haven't found a general way to reference fields/variables in C that overlap with reserved keywords in Python/Cython. Ultimately though, I don't think this is much of an issue for zfp.  Cython does not require that every field in a struct be listed.  It only requires those fields that you need to directly access from Cython code.  So we can just omit `exec` from the `zfp_stream` definition in Cython and then set `exec` via the `zfp_stream_set_execution` function.  If we need to reference the `exec` field directly from Cython/Python for some reason, we might need to change the name in C.

lindstro:
@SteVwonder I'm fine with that for the current Python implementation.  I still think we should take this into consideration and change C struct member names for the next release to avoid issues like this.

With regards to the `type` field, is it possible that the behavior differs between Python 2 and 3, or did you test both?

SteVwonder:
Just pushed a commit adding support for fixed-rate and fixed-precision modes.  The `compress_numpy` function now has two additional "optional" arguments.  Exactly one of the "optional" arguments (i.e., rate, precision, tolerance) must be provided by the user, otherwise an exception is thrown.  This eliminates the need for having three separate functions, but it is a bit odd to require an "optional" argument.  As a quick sanity check, is `-1` a valid value for rate or tolerance? (I'm currently using `-1` as the default value for all three optional arguments under the assumption its an invalid value).

Hopefully this works for the EOY milestone, and we can come up with a better interface for the March release.  Also, same caveat as before that the current python test cases just verify that nothing catastrophic happens when doing a round-trip compression/decompression using the bindings.

> I've been meaning to refactor some of those utility libraries, so your input can help re-shape those APIs to be the least painful when calling from Python (ex. removing double pointers).

After peeking at the utility libraries, I think I have a few ideas on how to make them more usable from Python. @salasoom, can we meet in person after the holidays to go over these tests a bit more in-depth and brainstorm?

> Can you base the PR from the develop branch?
> For now, please separate the tests into dir tests/python

👍 Done.

----

> I don't think we should have a default tolerance. The user should be aware that this is something they need to specify. If they don't know how to set this, then I think it would be safer to use a tolerance of zero.

Makes sense.  Removed the default tolerance as part of the fixed-rate/precision commit.

> Another difference wrt. zfp_compress is that only one field can be compressed at a time.

That's a good point.  I can document that as a limitation of the current numpy-specific interface.  Would adding a boolean `output_header` argument that toggles the writing of the header make it compatible with files compressed via the C interface?

> Does Cython expose internal implementation functions like _init_field in the global namespace? That would violate the ANSI standard, which reserves all public symbols starting with _.

`__init__` doesn't get exported as a symbol in the shared library, but other functions that start with _ are exported, like `_PyBytes_Type` and `___pyx_module_is_main_zfp`.  I don't think this should be an issue since only Python will be `dlopen`ing the shared library, but maybe I'm missing something.

> class Memory(): Maybe check the return value of malloc?
> The return value to zfp_decompress is not checked. It will be zero in case of failure. Maybe throw an exception?

Good ideas.  Both are done.

---

>  I still think we should take this into consideration and change C struct member names for the next release to avoid issues like this.

👍 sounds good to me.  Also, I spoke too soon. It does appear that Cython has support for struct members whose names conflict with python keywords [(cython reference)](http://docs.cython.org/en/latest/src/userguide/external_C_code.html#resolving-naming-conflicts-c-name-specifications). I believe the follow should work (will try out tomorrow).
```
ctypedef struct zfp_stream:
    zfp_exec _exec "exec"
```

> With regards to the type field, is it possible that the behavior differs between Python 2 and 3, or did you test both?

That is possible.  I have only been able to test against python 3 while on my mac. I can test against both py2 & 3 on Linux tomorrow.

lindstro:
@SteVwonder Well done!  I have a few comments below.

> Just pushed a commit adding support for fixed-rate and fixed-precision modes. The `compress_numpy` function now has two additional "optional" arguments. Exactly one of the "optional" arguments (i.e., rate, precision, tolerance) must be provided by the user, otherwise an exception is thrown. This eliminates the need for having three separate functions, but it is a bit odd to require an "optional" argument. As a quick sanity check, is `-1` a valid value for rate or tolerance? (I'm currently using `-1` as the default value for all three optional arguments under the assumption its an invalid value).

Thanks--I think this is fine for now.  -1 is not valid for any of the three parameters, so that'll work.  But I agree that we need a cleaner interface for the release.

> 
> Hopefully this works for the EOY milestone, and we can come up with a better interface for the March release. Also, same caveat as before that the current python test cases just verify that nothing catastrophic happens when doing a round-trip compression/decompression using the bindings.

OK.  I'm still concerned about the memory layout of the numpy array.  Did you get a chance to look into using strides?  I believe asfortranarray actually permutes the array, which could be expensive and unnecessary.  Instead, I would suggest mapping shape[3] to nx, shape[2] to ny, and so on.  This corresponds to the default ndarray layout and presumably the most common case.  Then set the zfp strides to guard against other layouts.  This ensures that in the most common case, we will traverse the numpy array in sequential order, and no deep copies are needed.

> 
> > I've been meaning to refactor some of those utility libraries, so your input can help re-shape those APIs to be the least painful when calling from Python (ex. removing double pointers).
> 
> After peeking at the utility libraries, I think I have a few ideas on how to make them more usable from Python. @salasoom, can we meet in person after the holidays to go over these tests a bit more in-depth and brainstorm?
> 
> > Can you base the PR from the develop branch?
> > For now, please separate the tests into dir tests/python
> 
> +1 Done.
> 
> > I don't think we should have a default tolerance. The user should be aware that this is something they need to specify. If they don't know how to set this, then I think it would be safer to use a tolerance of zero.
> 
> Makes sense. Removed the default tolerance as part of the fixed-rate/precision commit.
> 
> > Another difference wrt. zfp_compress is that only one field can be compressed at a time.
> 
> That's a good point. I can document that as a limitation of the current numpy-specific interface. Would adding a boolean `output_header` argument that toggles the writing of the header make it compatible with files compressed via the C interface?

Yes, I think that makes sense.

> 
> > Does Cython expose internal implementation functions like _init_field in the global namespace? That would violate the ANSI standard, which reserves all public symbols starting with _.
> 
> `__init__` doesn't get exported as a symbol in the shared library, but other functions that start with _ are exported, like `_PyBytes_Type` and `___pyx_module_is_main_zfp`. I don't think this should be an issue since only Python will be `dlopen`ing the shared library, but maybe I'm missing something.

That's fine.  I was just concerned about polluting the global namespace with internal functions and doing so in an ANSI non-compliant manner.

> 
> > class Memory(): Maybe check the return value of malloc?
> > The return value to zfp_decompress is not checked. It will be zero in case of failure. Maybe throw an exception?
> 
> Good ideas. Both are done.
> 
> > I still think we should take this into consideration and change C struct member names for the next release to avoid issues like this.
> 
> +1 sounds good to me. Also, I spoke too soon. It does appear that Cython has support for struct members whose names conflict with python keywords [(cython reference)](http://docs.cython.org/en/latest/src/userguide/external_C_code.html#resolving-naming-conflicts-c-name-specifications). I believe the follow should work (will try out tomorrow).

Nice.

> 
> ```
> ctypedef struct zfp_stream:
>     zfp_exec _exec "exec"
> ```
> 
> > With regards to the type field, is it possible that the behavior differs between Python 2 and 3, or did you test both?
> 
> That is possible. I have only been able to test against python 3 while on my mac. I can test against both py2 & 3 on Linux tomorrow.

Sounds good.  I noticed that this PR does not pass on Travis.  Do you know why?

SteVwonder:
> Did you get a chance to look into using strides?

I have not, but I agree that strides is definitely the cleaner/better way to go in the long term. Unfortunately, I will not be able to implement that before the EOY.

> I noticed that this PR does not pass on Travis. Do you know why?

I'm not 100% sure of the root cause, but I think it has to do with either the version of python (v3.4) or Cython (v?) in Ubuntu Trusty.  After switching to Ubuntu Xenial, everything passes.  I will investigate that further.

> Regarding the changes to the LICENSE, LLNL has gotten picky about what can go in this file: see software.llnl.gov/about/licenses. We may have to update the license and place this additional information in the python directory.

Forgot to mention earlier that I made this change.  The top-level LICENSE is unchanged, and there are two subdirectories under the `python` directory which contain copied code along with a copy of the original licenses for that code.

> > That's a good point. I can document that as a limitation of the current numpy-specific interface. Would adding a boolean output_header argument that toggles the writing of the header make it compatible with files compressed via the C interface?

> Yes, I think that makes sense.

Is there a file currently generated as part of the testsuite to test this against?

PS - I added a commit that removes references to reserved keywords like `bytes` and `type`.

lindstro:
> > Did you get a chance to look into using strides?
> 
> I have not, but I agree that strides is definitely the cleaner/better way to go in the long term. Unfortunately, I will not be able to implement that before the EOY.

No worries.  Let's tackle that next year.  Would be nice to verify that the Python version can decompress a file written by the zfp command-line tool (see more below).

> > I noticed that this PR does not pass on Travis. Do you know why?
> 
> I'm not 100% sure of the root cause, but I think it has to do with either the version of python (v3.4) or Cython (v?) in Ubuntu Trusty. After switching to Ubuntu Xenial, everything passes. I will investigate that further.

Nice!

> > Regarding the changes to the LICENSE, LLNL has gotten picky about what can go in this file: see software.llnl.gov/about/licenses. We may have to update the license and place this additional information in the python directory.
> 
> Forgot to mention earlier that I made this change. The top-level LICENSE is unchanged, and there are two subdirectories under the `python` directory which contain copied code along with a copy of the original licenses for that code.

Sounds good.

> > > That's a good point. I can document that as a limitation of the current numpy-specific interface. Would adding a boolean output_header argument that toggles the writing of the header make it compatible with files compressed via the C interface?
> 
> > Yes, I think that makes sense.
> 
> Is there a file currently generated as part of the testsuite to test this against?

No, we don't include binaries with zfp.  But, what I would suggest is that you generate a small rectangular 2D or 3D floating-point array and compress it with the zfp command-line tool (use -h to write a header).  A linear sequence of integers will do: f(x, y, z) = x + nx * (y + ny * z).  You want to make sure nx != ny != nz, e.g., nx = 15, ny = 16, nz = 17.  Then feed the compressed byte stream to your decompressor and verify that it works correctly.  You could use -a 0 to achieve near lossless compression.

Then repeat with zfPy compressing the file and the command-line tool decompressing.

> PS - I added a commit that removes references to reserved keywords like `bytes` and `type`.

Can you please elaborate?  Surely zfp_field.type needs to be referenced somewhere?

We should be careful to avoid Python keywords in the zfp C implementation.  I tried to locate a list of reserved keywords but did not see either `type` or `bytes` referenced.  Are these provided by Python modules?

SteVwonder:
> No, we don't include binaries with zfp. But, what I would suggest is that you generate a small rectangular 2D or 3D floating-point array and compress it with the zfp command-line tool (use -h to write a header). A linear sequence of integers will do: f(x, y, z) = x + nx * (y + ny * z). You want to make sure nx != ny != nz, e.g., nx = 15, ny = 16, nz = 17. Then feed the compressed byte stream to your decompressor and verify that it works correctly. You could use -a 0 to achieve near lossless compression.
> Then repeat with zfPy compressing the file and the command-line tool decompressing.

`zfPy`: I like that stylization 😄 

👍 Sounds like a solid plan. We should be able to write the "compress with zfPy and decompress with CLI" test after adding the `write_header` argument to `compress_numpy`.  I think we'll need to expose `zfp_field` as a "first-class citizen" before being able to write the reverse test, since that requires passing metadata about the array (i.e., type & dimensonality) to the decompression function.

> We should be careful to avoid Python keywords in the zfp C implementation. I tried to locate a list of reserved keywords but did not see either `type` or `bytes` referenced. Are these provided by Python modules?

Sorry, I was being sloppy with my terms.  `type` and `bytes` are technically built-in functions.  You can find the full list of built-ins [here](https://docs.python.org/3.7/library/functions.html#built-in-functions). As far as reserved keywords, the best reference I could find was in the [Cython source code](https://github.com/cython/cython/blob/2bb984a4f6c20687f50055e71a66813579b30485/Cython/Compiler/Scanning.py#L40).

> 
> > PS - I added a commit that removes references to reserved keywords like `bytes` and `type`.
> 
> Can you please elaborate? Surely zfp_field.type needs to be referenced somewhere?

You are absolutely correct; it does need to be referenced.  In the case of `type`, I used a feature of Cython to "rename" the struct member to `_type`.  The struct definition now looks like:
```
ctypedef struct zfp_field:
    zfp_type _type "type"
    ...
```
And the references to the member look like: `cdef zfp_type ztype = field[0]._type`.  I didn't actually run into any issues personally with `type`, other than syntax highlighting oddities, but out of an abundance of caution (paranoia?), I made the switch.  Since it wasn't actually an issue, I can revert the change if you want.  Since `exec` is actually a reserved keyword and not just a built-in function, we will need to do something similar to handle that case (at least until the definition in C changes).

In the case of `bytes`, it was only used in function definitions at the top of the file (i.e., `bitstream* stream_open(void* data, size_t bytes);`).  Cython actually only requires type information for the parameters, so I changed it to `bitstream* stream_open(void* data, size_t);`.

lindstro:
> Just pushed a commit adding support for fixed-rate and fixed-precision modes. The `compress_numpy` function now has two additional "optional" arguments. Exactly one of the "optional" arguments (i.e., rate, precision, tolerance) must be provided by the user, otherwise an exception is thrown. This eliminates the need for having three separate functions, but it is a bit odd to require an "optional" argument. As a quick sanity check, is `-1` a valid value for rate or tolerance? (I'm currently using `-1` as the default value for all three optional arguments under the assumption its an invalid value).

@SteVwonder We now have support for reversible (lossless) compression (see feature/lossless branch).  I think a quite natural default would be to use reversible compression when the user does not specify a rate, precision, or tolerance.  We'll have to work on integrating these branches, but I wanted to mention this as a perhaps better alternative than throwing an exception when no optional argument is passed.

rabernat:
Following this work with interest.

I have a question that I couldn't quite answer from reading the code: can the `decompress_numpy` decompress directly into a pre-allocated buffer? This would be useful for https://github.com/zarr-developers/numcodecs/issues/117.

python/zfp.pyx
```diff
@@ -0,0 +1,242 @@
...
+    cdef bitstream* bstream = stream_open(comp_data_pointer, len(compressed_data))
+    cdef zfp_field* field = zfp_field_alloc()
+    cdef zfp_stream* stream = zfp_stream_open(bstream)
+
+    zfp_stream_rewind(stream)
+    zfp_read_header(stream, field, HEADER_FULL)
+
+    cdef zfp_type ztype = field[0]._type
+    if ztype == zfp_type_int32 or ztype == zfp_type_int64:
+        raise NotImplementedError("Integer types not supported")
```
rabernat: maybe `TypeError` would be more appropriate here?

lindstro: As it turns out, integer types are implemented.  This if statement should be removed.

lindstro:
> Following this work with interest.
> 
> I have a question that I couldn't quite answer from reading the code: can the `decompress_numpy` decompress directly into a pre-allocated buffer? This would be useful for [zarr-developers/numcodecs#117](https://github.com/zarr-developers/numcodecs/issues/117).

@SteVwonder can correct me if I'm wrong, but the current implementation allocates a buffer (ndarray).  The reason for this is that the caller of decompress_numpy does not necessarily know how large a buffer to allocate since the header information that stores array dimensions and scalar type is part of the compressed stream.

If this metadata were available elsewhere, then I believe there's no technical reason why we could not supplement the API with a function that also accepts the buffer to decompress into.  We'd want to ensure that the buffer is large enough to hold the decompressed data.  The zfp C implementation requires the caller to allocate the buffer anyway, which is made easier by separating the calls for parsing the header and decompressing the array.

@rabernat Can you point us to en example of how such functionality is being used in Zarr/numcodecs?

rabernat:
> @rabernat Can you point us to en example of how such functionality is being used in Zarr/numcodecs?

The python wrapper for blosc in numcodecs contain an optional `dest` argument:
https://github.com/zarr-developers/numcodecs/blob/master/numcodecs/blosc.pyx#L330

SteVwonder:
> @SteVwonder can correct me if I'm wrong, but the current implementation allocates a buffer (ndarray). The reason for this is that the caller of decompress_numpy does not necessarily know how large a buffer to allocate since the header information that stores array dimensions and scalar type is part of the compressed stream.

:+1: 

> If this metadata were available elsewhere, then I believe there's no technical reason why we could not supplement the API with a function that also accepts the buffer to decompress into. 

Agreed!

lindstro:
> > @rabernat Can you point us to en example of how such functionality is being used in Zarr/numcodecs?
> 
> The python wrapper for blosc in numcodecs contain an optional `dest` argument:
> https://github.com/zarr-developers/numcodecs/blob/master/numcodecs/blosc.pyx#L330

Seems like a reasonable solution.  We should be able to do something similar.

zfp supports parallel compression using OpenMP and CUDA.  Is that something that would be useful in Zarr?

rabernat:
> zfp supports parallel compression using OpenMP and CUDA. Is that something that would be useful in Zarr?

Definitely!

I'm a bit concerned that we now have two competing implementations of the python bindings for zfp, the one here and the one from @halehawk in https://github.com/zarr-developers/numcodecs/pull/160.

halehawk:
I think it is fine. I implemented basic compression/decompression of zfp
which can be used immediately. Peter and his team can add parallel and
image compression slowly.

romankarlstetter:
Nice to see work on python bindings for zfp.

I would like to test the python bindings for zfp. How would I do this? As far as I see it, there is no release of this yet. How can I install and use the current development version of these bindings?

Thanks in advance :)

lindstro:
@romankarlstetter You can clone this repo: https://github.com/SteVwonder/zfp/tree/feature/py-bindings.  This is still very much work in development, and we expect further changes before the scheduled late March release, including support for lossless compression.  But this should get you going.  Any feedback would be appreciated.

SteVwonder:
> As far as I see it, there is no release of this yet.

@romankarlstetter: yeah, we are planning to get the bindings in good enough shape to be merged in for the March release.

> How can I install and use the current development version of these bindings?

You will have to clone my fork and checkout the `py-bindings` branch: https://github.com/SteVwonder/zfp/tree/feature%2Fpy-bindings.  Then when you run cmake, add `-DBUILD_PTYHON=on` to the line.

Just a heads up that we have tested a round-trip of compression/decompression with the bindings, but we have not tested compressing with the C/C++ APIs and then decompressing with the Python bindings (or vice versa).  So any existing compressed data that you have on disk may not be readable by the bindings.

> Thanks in advance :)

No problem.  Thanks for trying it out. Let us know if you have any issues.

cgohlke:
Thank you for providing this interesting library! 
FWIW, I added a ZFP codec to the [imagecodecs](https://pypi.org/project/imagecodecs/) Python package to experiment with ZFP compression in TIFF. Compared to this PR, imagecodecs's Cython bindings use `zfp_field_set_stride_`, release the GIL, can write to user provided output buffers, does more error checking, and allows to specify the execution model. See the [`_zfp.pyx`](https://www.lfd.uci.edu/~gohlke/code/_zfp.pyx.html) Cython file in the sdist file.

lindstro:
@cgohlke Thanks for the pointer to imagecodecs.  It seems like you've done a lot of work.  Do you mind if I add a link to imagecodecs off the [zfp website](https://computation.llnl.gov/projects/floating-point-compression/related-projects)?

Our own Python bindings are a fairly recent endeavor that we hope to develop fully over the next year.  For instance, we plan on adding bindings to zfp's compressed array classes that support random access to individual array elements, probably using a numpy interface.  We'll also add proper handling of strides and execution policies as well as mechanisms for specifying which dimensions in a multidimensional array are correlated.

I'm now aware of at least four independently developed zfp Python bindings.  With so much interest, we're committed to releasing and maintaining our own Python implementation that will incorporate new features of zfp as they are released.  We certainly do welcome upstream contributions to the benefit of all, so please consider submitting PRs and suggestions for improvement.

cgohlke:
> add a link to imagecodecs off the zfp website 

Please do.

> suggestions for improvement.

Releasing the Python GIL would be one. 

Probably I am missing something, but for me `zfp_compress` segfaults or writes an incomplete stream when the output buffer/stream is not long enough. I would expect the function to return 0 in those cases.

romankarlstetter:
> 
> 
> Any feedback would be appreciated.

First of all, thanks for the quick response. And the most important message first: **it works**.
I tried compressing/decompressing numpy arrays in a jupyter notebook under python 3.6.8

There are a few things that, when improved, would enhance the experience inside notebooks a lot.
I understand that, at this point, documentation is not a priority, but having at least some documentation would definitely make it easier. My suggestions:

- Use `-DCYTHON_FLAGS=-Xembedsignature=True,binding=True` by default for building (or specify it in the pyx source) for the functions which should be part of the public API (see [Cython compiler directives](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives))
 - disable the `--no-docstrings` option for release builds, which is set in python/scikit-build-cmake/UseCython.cmake so that docstrings are also available in release mode (default)

Regarding the API: consider using a `mode` parameter which specifies which mode to use (one of precision, rate, accuracy, expert, as string or enum, maybe with a good default (accuracy?)). And then use one of the kw-args, (depending on the mode) to further configure compression settings.

lindstro:
@romankarlstetter Thanks for your feedback and for verifying that the code works as expected.  We will add more extensive tests over the next month, as well as documentation.  For now, your best bet is https://zfp.readthedocs.io/en/release0.5.4/, which documents the C API in detail.

I will defer to @SteVwonder with regards to your Python specific suggestions.

markcmiller86:
> @SteVwonder We now have support for reversible (lossless) compression (see feature/lossless branch). I think a quite natural default would be to use reversible compression when the user does not specify a rate, precision, or tolerance. We'll have to work on integrating these branches, but I wanted to mention this as a perhaps better alternative than throwing an exception when no optional argument is passed.

Just wondering about this choice and how numpy+compression really winds up looking to an end user. I mean, why does a numpy user even decide they want to use compression? If they're looking at memory usage, its probably in ``top``, right?  I mean, thats the first place I'd look. Don't like how much memory you're using? Ok, make a quick switch to ZFP arrays because those compress. Look at it again in ``top``? Will lossless ZFP mode result in noticeable reductions? Note, this isn't a critical question about ZFP or lossless mode. Its just a question...is their likely to be a noticeable memory reduction? If not, user may, without looking further, decide to try something else? OTOH, if default for ZFP is a decent lossy mode, then in this simple example, user would be encouraged by what ``top`` says in the memory savings. Then, maybe they start looking at more of the ZFP details, maybe their computations wind up looking not as good, whatever. But their first sanity check (e.g. what ``top`` is telling them) suggests looking further.

python/zfp.pyx
SteVwonder: 👍 Integer support isn't 100% complete since I'm still working on getting integers to compress such that they match the known-good checksums, but this statement has been removed.

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+    elif ndim == 2:
+        field = zfp_field_2d(pointer, ztype, shape[0], shape[1])
+        zfp_field_set_stride_2d(field, strides[0], strides[1])
+    elif ndim == 3:
+        field = zfp_field_3d(pointer, ztype, shape[0], shape[1], shape[2])
+        zfp_field_set_stride_3d(field, strides[0], strides[1], strides[2])
+    elif ndim == 4:
+        field = zfp_field_4d(pointer, ztype, shape[0], shape[1], shape[2], shape[3])
+        zfp_field_set_stride_4d(field, strides[0], strides[1], strides[2], strides[3])
+    else:
```
lindstro: If the numpy array uses C order, which is more likely than not, then we'd expect strides[0] to be largest and strides[3] to be smallest.  zfp compression will traverse memory using strides[0] first, then strides[1], etc., which is likely to result in poor locality of reference.  When the array is C contiguous, we ought to map shape[3] and strides[3] to x, shape[2] and strides[2] to y, etc.

The above works but may not be efficient.

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+    except KeyError:
+        # TODO: is this the correct type?
+        print "Using np.bytes_"
+        return np.bytes_
+
+cdef size_t sizeof_ztype(zfp_type ztype):
+    if ztype == zfp_type_int32:
+        return sizeof(signed int)
+    elif ztype == zfp_type_int64:
+        return sizeof(signed long long)
```
lindstro: Not sure if Cython assumes any particular C standard, but long long is not C89.  We use some preprocessor magic to provide int32 and int64 types in zfp.  Can those be used instead?

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+        raise ValueError("Only one of tolerance, rate, or precision can be set")
+
+    # Setup zfp structs to begin compression
+    cdef zfp_field* field = _init_field(arr)
+    stream = zfp_stream_open(NULL)
+
+    cdef zfp_type ztype = zfp_type_none;
+    cdef int ndim = arr.ndim;
+    if tolerance > -1:
+        zfp_stream_set_accuracy(stream, tolerance)
```
lindstro: This still allows for potential user error, e.g., if the user passes -0.5 as the tolerance.  tolerance, rate, and precision all have to be non-negative.

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+    cdef bytes compress_str = None
+    cdef size_t maxsize = zfp_stream_maximum_size(stream, field)
+    with Memory(maxsize) as data:
+        bstream = stream_open(data, maxsize)
+        zfp_stream_set_bit_stream(stream, bstream)
+        zfp_stream_rewind(stream)
+        # write the full header so we can reconstruct the numpy array on
+        # decompression
+        if write_header:
+            zfp_write_header(stream, field, HEADER_FULL)
```
lindstro: For safety, I'd suggest checking the return value to ensure it is not zero.

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+    if compressed_data is None:
+        raise TypeError
+
+    cdef char* comp_data_pointer = compressed_data
+    cdef bitstream* bstream = stream_open(comp_data_pointer, len(compressed_data))
+    cdef zfp_field* field = zfp_field_alloc()
+    cdef zfp_stream* stream = zfp_stream_open(bstream)
+
+    zfp_stream_rewind(stream)
+    zfp_read_header(stream, field, HEADER_FULL)
```
lindstro: This assumes that compress_numpy wrote a header or that the compressed data came from somewhere else.  At the very least, the return value should be checked.  If no header (or less than a full header) was stored, then this function will fail and return zero.

The documentation needs to make it clear that decompress_numpy requires a header.

python/zfp.pyx
```diff
@@ -0,0 +1,210 @@
...
+
+    cdef zfp_type ztype = field[0]._type
+
+    output_arr = _decompress_with_view(field, stream)
+
+    zfp_field_free(field)
+    zfp_stream_close(stream)
+    stream_close(bstream)
+
+    return output_arr
```
lindstro: zfp_decompress returns the current bit offset after decompression.  For decompressing single arrays, we don't need that info, but if we want to support decompressing several arrays stored in the same stream, then we need a mechanism to determine how many compressed bytes were consumed.

We might want to point out differences like this wrt. the C API in the documentation.

python/zfp.pyx
SteVwonder: :+1: It turns out that the whole function can be replaced with `numpy.dtype`'s `itemsize` method.

python/zfp.pyx
SteVwonder: :+1: great point!  Updated to check for `>= 0`

python/zfp.pyx
SteVwonder: Done! It now throws a `RuntimeError` if the return is 0.

python/zfp.pyx
SteVwonder: In the process of supporting @salasoom's new test utils checksum functions, this whole function got overhauled.  The decompression now only requires either the header or the user to provide the metadata (compression parameters, shape, strides, and type).  If both are provided, then they are compared against one another and an exception is raised if they differ (e.g., the header has `float` as the type but the user specified `int32`).

python/zfp.pyx
SteVwonder: Do we want to (by default) do the same thing on the decompression-side too?  That would allow a user to do a "round-trip" through the compressor/decompressor and still have the same data layout at the end of it.

Also, should we try and analyze the numpy strides to determine if they should be flipped or not?  In the off-chance that the numpy array is in Fortran ordering, then the strides should not be reversed for optimal locality, right?

python/zfp.pyx
SteVwonder: :+1: Noted.  I'll leave this comment unresolved as a reminder until I actually address it in the documentation.

SteVwonder:
> I have a question that I couldn't quite answer from reading the code: can the decompress_numpy decompress directly into a pre-allocated buffer? This would be useful for zarr-developers/numcodecs#117.

@rabernat: just a heads up that the latest code in this PR now supports decompression into a user-provided buffer

SteVwonder:
> Compared to this PR, imagecodecs's Cython bindings use zfp_field_set_stride_, release the GIL, can write to user provided output buffers, does more error checking, and allows to specify the execution model.

@cgohlke: Thanks for the feedback. Just a heads up that the latest code in this PR set the strides, releases the GIL (on the calls to zfp_(de)compress), and improves error checking and input validation.   Specifying the execution model is a longer term milestone, so it won't be added for this release, but hopefully soon.

SteVwonder:
> There are a few things that, when improved, would enhance the experience inside notebooks a lot.
I understand that, at this point, documentation is not a priority, but having at least some documentation would definitely make it easier. My suggestions:
> - Use -DCYTHON_FLAGS=-Xembedsignature=True,binding=True by default for building (or specify it in the pyx source) for the functions which should be part of the public API (see Cython compiler directives)
> - disable the --no-docstrings option for release builds, which is set in python/scikit-build-cmake/UseCython.cmake so that docstrings are also available in release mode (default)

@romankarlstetter: Thanks for giving the bindings a spin and for the feedback.  I modified the cmake to add `-Xembedsignature=True` and not include `--no-docstrings` (except in the case of `CMAKE_BUILD_TYPE=MinSizeRel`).  It appears that `-Xbinding=True` has performance implications, so I opted to leave it out for now.   For computationally-intensive functions (like compress & decompress), it doesn't look like the overheads are bad, but I wanted to err on the performance-side here.   Do you know if `embedsignature` is enough to be useful in Jupyter Notebooks?  Or do you need the introspection enabled by `binding=True`?

> Regarding the API: consider using a mode parameter which specifies which mode to use (one of precision, rate, accuracy, expert, as string or enum, maybe with a good default (accuracy?)). And then use one of the kw-args, (depending on the mode) to further configure compression settings.

Per the default mode: That is a good suggestion.  In previous conversations with @lindstro, my impression is that it is hard to define a "good default" since it is so data/domain dependent.  Luckily, lossless compression is being added in this release, so I think we are moving towards @markcmiller86's suggestion of lossless being the default.

Per the kwargs suggestion: I'm open to changing the compression settings part of the API to be two arguments (mode + mode parameter) rather than three (precision, rate, tolerance).  One nice thing about the current API is that the the various compression parameters types (tolerance and rate are doubles while precision is an int) are statically defined and then checked at runtime automatically by Cython.  Not a big deal to do that checking manually though.

python/zfp.pyx
SteVwonder: To hopefully clarify my previous comment, I am thinking of the case where a user has a "C ordered" numpy array that they want to compress and then decompress.

If they do this round-trip with the current code, the ordering of elements in their decompressed array will match the order in their original array (but the compression will be suboptimal).

If they do this round-trip with the strides in the above code (which only runs on compression) flipped, then the compression will be much better but the ordering of the elements in the decompressed array will not match the ordering of the elements in the original array unless the user explicitly provides the `strides` argument to the decompression function.  This seems like surprising behaviour to me.  I would normally expect a round-trip through compression/decompression to not affect element ordering (or at minimum, for the interface to populate the numpy array with the proper metadata so that it can "lie" to me about the underlying data ordering and provide me with a "view" of the data whose order I expect.)

To get the best of both worlds, maybe we switch to analyzing the strides of the input numpy array so that it is compressed (by default) with the fastest changing dimension first.  Then on decompression, we (by default) decompress assuming that the fastest changing dimension is first.  We can either decompress into "C order" (which seems to be numpy's default/preferred order) or we could decompress in "F order" and set the numpy metadata such that it is aware of this ordering (and then can adjust the view it provides the user depending on what they want).

Hopefully that makes more sense. I may be too close to the problem though and getting myself turned around.  Maybe a face-to-face meeting would help clear things up.

lindstro: I'm confused by your "compression will be suboptimal" comment.  Do you mean that the compression *speed* (not compression ratio) will be suboptimal?  We definitely do need to make sure that numpy and zfp array dimensions agree to get a good compression ratio.  The issue we're discussing here is about the memory layout, not the shape of the array.

Yes, it would be preferable to have the memory layouts agree between compression and decompression, but if the user passes a numpy array to decompress into, then we're at their mercy.  Moreover, nothing is stored in the zfp stream about the source memory layout.

One problem with "analyzing" the strides is that the user may not even be aware of the memory layout of their array.  They know the shape and would expect the same shape returned on decompression.  Note that this shape information may be passed as metadata to some other application, perhaps written in some other language.  If we silently permute the dimensions, then I fear things can go wrong.

I think it's better to just use the strides reported by the numpy array when compressing and to assume (for performance reasons) that the most common case is C order.

On decompression, we may want to allow the user to specify if they want their array in C or Fortran contiguous order.  Of course, the current API allows overriding that with strides.  I think we need to chat in person about how to deal with this.

romankarlstetter:
> 
> @romankarlstetter: Thanks for giving the bindings a spin and for the feedback. I modified the cmake to add `-Xembedsignature=True` and not include `--no-docstrings` (except in the case of `CMAKE_BUILD_TYPE=MinSizeRel`). It appears that `-Xbinding=True` has performance implications, so I opted to leave it out for now. For computationally-intensive functions (like compress & decompress), it doesn't look like the overheads are bad, but I wanted to err on the performance-side here. Do you know if `embedsignature` is enough to be useful in Jupyter Notebooks? Or do you need the introspection enabled by `binding=True`?

I just tried that, `binding=True` is not necessary, the other two options are sufficient.

> Per the default mode: That is a good suggestion. In previous conversations with @lindstro, my impression is that it is hard to define a "good default" since it is so data/domain dependent. Luckily, lossless compression is being added in this release, so I think we are moving towards @markcmiller86's suggestion of lossless being the default.

That sounds great 🙂

salasoom:
@SteVwonder I've set up Travis CI to always build & test Python. It's using Python 3.5 by default, with some build jobs testing 2.7. (I saw a compiler error on 2.7 with Cython)

SteVwonder:
@salasoom: thanks!  I saw the Cython error last night.  It looks like the Cython in Travis is not happy when argument names are left out of function signatures (in this case `bytes` in `stream_open`). I'm wondering if it is related to the Cython version.  I'm currently using Spack to install several older versions of Cython in an attempt to reproduce the error.

salasoom:
@SteVwonder At one point I did see 2.7 pass on trusty, when I was trying a bunch of different configs to get Python CI working.
https://travis-ci.org/LLNL/zfp/jobs/511857066

That passing job used cython 0.23.4, while the failing 2.7 jobs use 0.20.1

python/zfp.pyx
```diff
@@ -0,0 +1,423 @@
...
+    int precision
+):
+    # check that the header and user type matches
+    if ztype != type_none and ztype is not field[0]._type:
+        raise ValueError(
+            "User-provided zfp_type does not match zfp_type in header"
+        )
+
+    # check that the header and user shapes match
+    header_shape = (field[0].nx, field[0].ny, field[0].nz, field[0].nw)
```
lindstro: Usually we map nx to the rightmost dimension of a numpy array.  Should these be reversed?

python/zfp.pyx
```diff
@@ -0,0 +1,423 @@
...
+        shape_dims = sum([1 for x in shape if x > 0])
+        if len(strides) != len(shape):
+            raise ValueError(
+                "Mis-match in shape and stride lengths"
+            )
+
+    # check that setting the compression parameters based on user input
+    # does not change the stream mode (i.e., the compression parameters
+    # provided by the user and in the header match)
+    if tolerance >= 0 or rate >= 0 or precision >= 0:
```
lindstro: I think we also need to handle the case when none of these are true, i.e., when reversible mode is requested.

python/zfp.pyx
```diff
@@ -0,0 +1,423 @@
...
+            zfp_stream_rewind(stream)
+            if ztype == type_none or shape is None:
+                raise ValueError(
+                    "Failed to read zfp header and the ztype/shape/mode "
+                    "were not provided"
+                )
+            zshape = gen_padded_int_list(shape, pad=0, length=4)
+            # set the shape, type, and compression mode
+            # strides are set further down
+            field[0].nx, field[0].ny, field[0].nz, field[0].nw = zshape
```
lindstro: I believe these ought to be reversed.

python/zfp.pyx
```diff
@@ -0,0 +1,423 @@
...
+                shape,
+                strides,
+                tolerance,
+                rate,
+                precision
+            )
+
+        # pad the shape with zeros to reach len == 4
+        strides = gen_padded_int_list(strides, pad=0, length=4)
+        field[0].sx, field[0].sy, field[0].sz, field[0].sw = strides
```
lindstro: Should these be reversed?

python/zfp.pyx
```diff
@@ -0,0 +1,423 @@
...
+
+        if out is None:
+            output = np.asarray(_decompress_with_view(field, stream))
+        else:
+            if isinstance(out, np.ndarray):
+                output = out
+            else:
+                header_dtype = ztype_to_dtype(field[0]._type)
+                header_shape = (field[0].nx, field[0].ny, field[0].nz, field[0].nw)
```
lindstro: Should these be reversed?

python/zfp.pyx
SteVwonder: This is a good point. We may have to rethink the API for `_decompress`.  How to differentiate between the user not specifying anything and the user intended to specify `reversible`?  Currently, there is ambiguity.

In the case where the user doesn't want to specify anything, we should check against the header.  In the case where the user does want to specify reversible, we should check.  Maybe we just add an additional default argument `reversible = False` and if the other three are `-1` and `reversible == False`, then don't check, else check the parameter that is set.

python/zfp.pyx
SteVwonder: This has been resolved by reversing the shape before the padding (line 372 in this preview).

python/zfp.pyx
SteVwonder: :+1: This has been resolved by reversing the strides before the padding (line 395 in this preview).

SteVwonder:
I believe there are still some issues with the handling of the `strides` argument to `_decompress`, but since that interface is still "experimental", I don't think that should block this PR.

python/zfp.pyx
lindstro: From a use case standpoint, we'd ideally want the user to specify array metadata and compression parameters only when a header is not present.  When the header is present, I don't think we even want to allow the user to specify parameters.  All we could do then is verify that those parameters match the header.  If somehow the user already knows this information, then it made little sense to write it in a header in the first place.

Would it make sense to add a Boolean, read_header, that is the counterpart to the write_header Boolean specified in the compressor?  If it is true, then attempt to read the header and fail if not present.  If false, then assume there's no header and use the user-supplied compression parameters

Alternatively, instead of a Boolean argument, this could also be separated into two different functions (one that requires a header and one that forbids a header).  This would help avoid incompatible arguments.

I prefer this over adding another potentially incompatible parameter to specify that the user expects reversible compression.

lindstro:
@SteVwonder Thanks for quick fixes and for pointing out the `strides` issue.  We can live with this for now with proper documentation.

python/zfp.pyx
SteVwonder: > Alternatively, instead of a Boolean argument, this could also be separated into two different functions (one that requires a header and one that forbids a header). This would help avoid incompatible arguments.

> I prefer this over adding another potentially incompatible parameter to specify that the user expects reversible compression.

Sure.  We already have a function that requires a header (`decompress_numpy`).  We could then change `_decompress` to forbid a header.  This will greatly simplify the `_decompress` function since it will no longer have two "personalities".  Do this sound good to you @lindstro?

If we go this route, should `_decompress` still attempt to read a header and throw an exception if it detects one?  Or should it just trust that the user provided a stream with no header?  I would normally lean towards extra checking, but I'm a little concerned about a *false positive* header detection (i.e., by chance the first bits of the headerless stream decode to `zfp`).  Thoughts?

lindstro: Great point!  Yes, let's simplify `_decompress` so that it does not read a header.  Since this function is not (yet) officially part of the API and is intended for experts only, I think it's fine to require the user to know what they're doing.

One downside of this design is that there is no way to first compress via `compress_numpy`, which outputs a header, and later decompress into a user-allocated buffer via `_decompress`, which now expects no header.  This suggests the need for a corresponding `_compress` function that does not write a header.

SteVwonder: > Great point! Yes, let's simplify _decompress so that it does not read a header. Since this function is not (yet) officially part of the API and is intended for experts only, I think it's fine to require the user to know what they're doing.

Sounds good. I'll make this change now.

> One downside of this design is that there is no way to first compress via compress_numpy, which outputs a header, and later decompress into a user-allocated buffer via _decompress, which now expects no header. This suggests the need for a corresponding _compress function that does not write a header.

The `compress_numpy` function has an optional `write_header` boolean argument to enable this use-case (as well as repeated compression to the same stream). The checksum tests make use of the `write_header` flag: https://github.com/LLNL/zfp/pull/29/files#diff-9fbf217f1e38984b823dcd123ec51ec8R64

lindstro: Of course.  Not sure what I was thinking.  Thanks for taking care of this issue.

SteVwonder: No worries!  While I'm refactoring the functions, do you mind if I remove `strides` as an argument to `_decompress`?  `strides` for decompression is currently broken in ways that I don't fully understand.  I think it may be better to withhold that functionality for now, and add it in a later release after it is tested and fully working. Thoughts?

lindstro: I can live with that.  Does this mean that strides will be inferred from the numpy array, when that's the target, or will otherwise be set to default values (sx = 1, sy = nx, sz = nx * ny, sw = nx * ny * nz)?

PS. One issue with strides that I pointed out earlier is that zfp measures strides in scalars while ndarray measures strides in bytes.

SteVwonder: > Does this mean that strides will be inferred from the numpy array, when that's the target, or will otherwise be set to default values (sx = 1, sy = nx, sz = nx * ny, sw = nx * ny * nz)?

I was thinking of leaving them unspecified (i.e., not modifying the stride values in the `field` so that they all remain 0 from initialization) and letting zfp calculate them.  Does that work?

PS - since the header is now assume to not exist, `shape` and `ztype` are no longer optional arguments to `_decompress`, they are now required arguments.

lindstro: > > Does this mean that strides will be inferred from the numpy array, when that's the target, or will otherwise be set to default values (sx = 1, sy = nx, sz = nx * ny, sw = nx * ny * nz)?
> 
> I was thinking of leaving them unspecified (i.e., not modifying the stride values in the `field` so that they all remain 0 from initialization) and letting zfp calculate them. Does that work?

I'm not opposed to that, but what happens if the user passes an array to decompress into?  Its strides will be ignored and not updated, right?  If those strides do not match what zfp assumes, then things will go wrong.  

> PS - since the header is now assume to not exist, `shape` and `ztype` are no longer optional arguments to `_decompress`, they are now required arguments.

Makes sense.  What happens if they do not match `out` when the user passes an array?

SteVwonder: > I'm not opposed to that, but what happens if the user passes an array to decompress into? Its strides will be ignored and not updated, right? If those strides do not match what zfp assumes, then things will go wrong.

Correct.  The user will have to do some massaging of the data/strides outside of ZFP to get things working if they don't want the default ZFP strides.

> Makes sense. What happens if they do not match out when the user passes an array?

Per our offline discussion, I'll refactor so that we continue to check that the out shape and type match if it is a numpy array, but also document that the user can pass in `array.data` as an "escape hatch" that avoids the check if they are confident they want to avoid it.

SteVwonder: This thread should be resolved as of the latest commit.

SteVwonder:
I think the AppVeyor failure is a false negative.  Only one builder failed, and within that builder, no tests failed.  The submission to cdash timed out:

```
Submit files (using https)
   Send to track: AppVeyor
   Using HTTP submit method
   Drop site:https://open.cdash.org/submit.php?project=zfp
   Error when uploading file: C:/projects/zfp/build/Testing/20190409-0345/Test.xml
   Error message was: Failed to connect to open.cdash.org port 443: Timed out
   Problems when submitting via HTTP
Command exited with code 127
```

PR #33, opened 2018-12-14
Use https:// instead of git:// for cryptomilk

mohamed:
In many companies, the git:// protocol is blocked by corporate firewall. To circumvent that, use http:// instead of git://

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -10,7 +10,7 @@ list(APPEND CMOCKA_ARGS "-DWITH_STATIC_LIB=ON;-DCMAKE_C_COMPILER=${CMAKE_C_COMPI
 include(ExternalProject)
 ExternalProject_Add(
   cmocka_cloned
-  GIT_REPOSITORY    git://git.cryptomilk.org/projects/cmocka.git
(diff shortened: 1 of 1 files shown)


mohamed:
@salasoom @lindstro BTW, I have just tested "https://" instead of "http://" and that seems to work as well. I leave it up to you guys to decide which protocol to use.

lindstro:
@mohamed Interesting.  We have the exact reverse problem at LLNL, where pushes over https are blocked, and we have to use git+ssh.  But I see no particular issue with changing the protocol for cmocka.

@salasoom Do you agree?

mohamed:
@lindstro You don't need to push over https. This is simply pulling from the cmocka repo. I hope that LLNL doesn't block read-only https access.

lindstro:
@mohamed Right.  I was merely pointing out that our default protocol is git+ssh so we can push new zfp code to GitHub.  That's probably why we're not using https for pulling cmocka, but as you note, we should be able to do so without problem.

salasoom:
We can go with https here. I will rebase the branch against develop, and amend the commit to use https.

PR #35, opened 2018-12-14
Unroll deposit bit plane loop

LennartNoordsij:
Setting the bound static prevents Warp divergence and speeds up the writing.
CUDA 8 supports specifying the unrolling factor which is done in other functions of the CUDA implementation as well.

### DIFF
--- a/src/cuda_zfp/decode.cuh
+++ b/src/cuda_zfp/decode.cuh
@@ -127,8 +127,12 @@ void decode_ints(BlockReader<Size> &reader, uint &max_bits, UInt *data)
       for (; n < (Size - 1) && bits && (bits--, !reader.read_bit()); n++);
     
     // deposit bit plane
+#if (CUDART_VERSION < 8000)
     #pragma unroll
-    for (int i = 0; x; i++, x >>= 1)
+#else
+    #pragma unroll Size
+#endif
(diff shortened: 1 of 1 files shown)


mclarsen:
Thanks!

lindstro:
@LennartNoordsij This sounds a bit like a related optimization that we have been working on.  One of the steps of zfp is transposition of the binary matrix of coefficient bits so that the matrix is organized by column (bit plane) instead of by row (coefficient).  In the current CPU implementation, this transposition is done on the fly by repeated access to the rows to extract each column, i.e., for precision p and dimensionality d, we make 4^d * p memory accesses.  Using a more optimized explicit transposition, we reduce this to lg(4^d) * p = 2 * d * p memory accesses (more than 10x fewer in 3D).  Moreover, this computation is highly parallelizable as the algorithm is based on divide and conquer.

On the CPU, I've seen a 20x speedup of the transposition itself and a resulting 3x speedup of the (de)compressor using this optimization, which completely unrolls the whole loop over bit planes (the outer for (k) loop).  This would likely run even faster if we take advantage of parallelism, but would require a significant rewrite of the CUDA implementation, which as you know currently parallelizes only over zfp blocks.

Anyway, since it seems related to your optimization, I thought I'd mention it.

PR #38, opened 2019-01-09
OpenMP decompression

LennartNoordsij:
This patch adds support in zfp for OpenMP decompression.
It does that by generating an "offset table" during the compression
phase, and then using that table for performing parallel variable-rate
decoding. The patch breaks the following tests:

- testOmp: The test is broken because the behavior of chunk_size is
           changed. We need to align with the maintainers on what to
           do regarding that.

The patch has been tested locally for 1D, 2D, 3D (4D is not tested yet).
The tests will be added in a follow up once we align on the implementation.


include/zfp.h
```diff
@@ -145,6 +145,7 @@ typedef struct {
   int minexp;         /* minimum floating point bit plane number to store */
   bitstream* stream;  /* compressed bit stream */
   zfp_execution exec; /* execution policy and parameters */
+  unsigned long long * offset_table; /* table with the chunk offsets in bits */
```
mohamed: Why not size_t?

include/bitstream.h
```diff
@@ -81,6 +81,9 @@ size_t stream_flush(bitstream* stream);
 /* copy n bits from one bit stream to another */
 void stream_copy(bitstream* dst, bitstream* src, size_t n);
 
+/* return pointer to start of data array used by this bitstream */
+void * stream_begin(bitstream* s);
```
mohamed: Change the return type to `const void *`

src/share/parallel.c
```diff
@@ -83,4 +83,30 @@ compress_finish_par(zfp_stream* stream, bitstream** src, uint chunks)
...
 }
 
-#endif
+/* initialize per-thread bit streams for parallel decompression */
+static bitstream**
+decompress_init_par(zfp_stream* stream, const zfp_field* field, uint chunks)
+{
+  uint i;
+  void * buffer = stream_begin(zfp_stream_bit_stream(stream));
+  bitstream** bs = (bitstream**)malloc(chunks * sizeof(bitstream*));
```
mohamed: What happens if `malloc` fails? You are not checking if `bs` is not `NULL`

utils/zfp.c
```diff
@@ -531,8 +553,16 @@ int main(int argc, char* argv[])
         fprintf(stderr, "cannot create compressed file\n");
         return EXIT_FAILURE;
       }
+      /* TODO: think of a clean way to check for offset header, possibly execution policy */
+	    if (chunk_size) {
+        /* write the offset header to the file */
+        if (fwrite(offset_table, sizeof(unsigned long long), chunks, file) != chunks) {
```
mohamed: I think you need to align with @lindstro and @mclarsen on how to handle writing the offset table. Few questions that pop up:
- Should it go to the header or a separate file?
- If it goes to the header, do we version the header to preserve backwards-compatibility?

src/share/parallel.c
lindstro: Good point.  I believe this is the only place in zfp where malloc is not being checked.

utils/zfp.c
lindstro: This will break compatibility in the sense that existing code will not be able to read a file that stores these offsets.  We are planning on making several changes to the compressed format for our September release, and that would be the right time to address this.  I'm afraid we can't officially adopt changes like this until then.

Perhaps we can make an experimental branch for evaluating parallel decompression and experimenting with different ways of encoding the offset table.  For instance, I had a discussion with @LennartNoordsij about alternative solutions that are not based on storing offsets.

include/zfp.h
LennartNoordsij: Agreed, I will change this to be a fixed width type. uint64_t is good for now.

include/bitstream.h
LennartNoordsij: It is not different, I must have missed that function. I will remove this function and use stream_data.

src/share/parallel.c
LennartNoordsij: This malloc is taken from line 51 of compress_init_par, where it is also unchecked. I will include a check for this in my function.
I will also raise an issue for the unchecked mallocs in compress_init_par.

utils/zfp.c
LennartNoordsij: An experimental branch seems like the best way to go forward for now. I will also add my CUDA decompression implementation which is based on the same offset table implementation as this OpenMP version. I will combine them both in a branch "Parallel decompression".

LennartNoordsij:
Closed until the design objective and implementation are aligned

PR #40, opened 2019-01-10
Check for failing mallocs

LennartNoordsij:
malloc can fail and return a NULL pointer. Previously this was unchecked.
This commit adds a check for that.

### DIFF
--- a/src/share/parallel.c
+++ b/src/share/parallel.c
@@ -49,9 +49,17 @@ compress_init_par(zfp_stream* stream, const zfp_field* field, uint chunks, uint 
 
   /* set up buffer for each thread to compress to */
   bs = (bitstream**)malloc(chunks * sizeof(bitstream*));
+  if (!bs) {
+    fprintf(stderr, "cannot allocate memory for per-thread bit streams\n");
+    exit(EXIT_FAILURE);
+  }
   for (i = 0; i < chunks; i++) {
     uint block = chunk_offset(blocks, chunks, i);
     void* buffer = copy ? malloc(size) : (uchar*)stream_data(stream->stream) + stream_size(stream->stream) + block * stream->maxbits / CHAR_BIT;
+    if (copy && !buffer) {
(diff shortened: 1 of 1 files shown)


src/share/parallel.c
```diff
@@ -49,9 +49,17 @@ compress_init_par(zfp_stream* stream, const zfp_field* field, uint chunks, uint
 
   /* set up buffer for each thread to compress to */
   bs = (bitstream**)malloc(chunks * sizeof(bitstream*));
+  if (!bs) {
+    fprintf(stderr, "cannot allocate memory for per-thread bit streams\n");
+    exit(EXIT_FAILURE);
```
lindstro: I'm afraid that exiting the entire application on failure is not acceptable.  zfp also should not write to stderr.  We need to return NULL here and have the caller check the return value.

src/share/parallel.c
```diff
@@ -49,9 +49,17 @@ compress_init_par(zfp_stream* stream, const zfp_field* field, uint chunks, uint
...
+  }
   for (i = 0; i < chunks; i++) {
     uint block = chunk_offset(blocks, chunks, i);
     void* buffer = copy ? malloc(size) : (uchar*)stream_data(stream->stream) + stream_size(stream->stream) + block * stream->maxbits / CHAR_BIT;
+    if (copy && !buffer) {
+      fprintf(stderr, "cannot allocate memory for compression buffer\n");
+      exit(EXIT_FAILURE);
```
lindstro: Another case where exiting is not a viable option.  We need to clean up after ourselves and then return NULL.

lindstro:
@LennartNoordsij Thanks for pointing out this issue and for your contribution.  I fear that this PR requires additional work, since terminating the calling application when memory allocation fails is not a viable solution.  Doing this correctly requires additional code to free any partially allocated resources, reporting an error back to the caller, and having the caller take appropriate action.

I have pushed a revised version to the develop branch and am closing this PR.

PR #42, opened 2019-01-11
warnings: mark conditionally used parameters

mathstuf:


lindstro:
Thanks for this PR.  Can you please resubmit it on the develop branch?

mathstuf:
Updated.

lindstro:
> I've modified your changes to hide behind a macro which describes why the line is there. I prefer that over adding a short comment in each spot. Are you ok with this?

I think this is a good solution, but I would also `#undef` this macro at the end of file to avoid it from propagating further.  When we want these C functions to be inlined, this source file is included in user code.

Finally, to reduce the likelihood of name clashes, we might want to use the same convention we do for similar macros (e.g. `inline_`, `extern_`, `restrict_`, `align_`, etc.) and name this macro `unused_`.

mathstuf:
Maybe calling it `ZFP_UNUSED` would be better?

salasoom:
Good point about adding `#undef` and I agree we should follow the naming convention, so it will be called `unused_`

mathstuf:
Thanks.

PR #46, opened 2019-02-13
Travis CI builds now report to the zfp cdash instance

robertmaynard:
Travis builders will now report progress to the zfp cdash instance ( https://open.cdash.org/index.php?project=zfp ).

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -9,11 +9,11 @@ if [ -n "${COVERAGE}" ]; then
   # build
   ctest -V -C "Debug" -DC_STANDARD=${C_STANDARD:-99} -DCXX_STANDARD=${CXX_STANDARD:-98} -DBUILD_CFP=ON -DBUILD_OPENMP=ON -DBUILD_CUDA=OFF -DWITH_COVERAGE=ON -S $TRAVIS_BUILD_DIR/cmake/travis.cmake
 else
-  # build/test without OpenMP, with CFP
-  ctest -V -C "Debug" -DC_STANDARD=${C_STANDARD:-99}  -DCXX_STANDARD=${CXX_STANDARD:-98} -DBUILD_CFP=ON -DBUILD_OPENMP=OFF -DBUILD_CUDA=OFF -S $TRAVIS_BUILD_DIR/cmake/travis.cmake
+  # build/test without OpenMP, with CFP (and custom namespace)
+  ctest -V -C "Debug" -DC_STANDARD=${C_STANDARD:-99}  -DCXX_STANDARD=${CXX_STANDARD:-98} -DBUILD_CFP=ON -DCFP_NAMESPACE=cfp2 -DBUILD_OPENMP=OFF -DBUILD_CUDA=OFF -S $TRAVIS_BUILD_DIR/cmake/travis.cmake
 
   rm -rf ./* ;
 
-  # build/test with OpenMP, with CFP custom namespace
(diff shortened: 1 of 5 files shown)


lindstro:
@robertmaynard That was quick!  Thanks for your help.

@salasoom Can you please review and merge this PR when you get a chance?

robertmaynard:
@salasoom I will start working on appveyor and have it as a follow on merge request.

salasoom:
@robertmaynard On CDash I signed in with my github account but am seeing a message, "you don't have permission to access this page" when I follow that link.

robertmaynard:
@salasoom I have added you as a user so you can see the project now. We can make the project public so everyone can see it if you want.

lindstro:
Can you please add me as well?  We may well decide to make this public, but I want to learn more about CDash first.

robertmaynard:
@lindstro You now have access

travis.sh
```diff
@@ -4,25 +4,14 @@ set -e
...
-
+  ctest -V -C "Debug" -DC_STANDARD=${C_STANDARD:-99} -DCXX_STANDARD=${CXX_STANDARD:-98} -DBUILD_CFP=ON -DBUILD_OPENMP=ON -DBUILD_CUDA=OFF -DWITH_COVERAGE=ON -S $TRAVIS_BUILD_DIR/cmake/travis.cmake
 else
-
   # build/test without OpenMP, with CFP
-  cmake .. -DCMAKE_C_STANDARD=${C_STANDARD:-99} -DCMAKE_CXX_STANDARD=${CXX_STANDARD:-98} -DZFP_WITH_OPENMP=OFF -DBUILD_CFP=ON;
-  cmake --build . ;
-  ctest -V -C "Debug";
-
-  rm -rf ./* ;
```
salasoom: We create two builds per CI job because some tests are exclusive to builds with/without OpenMP. In this commit's travis log, it looks like the OpenMP tests are never compiled or run (named testZfpOmp*). I think the `rm -rf ./*` to clear the `build/` directory is missing.

Is CDash capable of handling multiple builds in 1 CI job? If we split this into separate jobs, then we double our CI build matrix.

robertmaynard: We can handle multiple builds in a single CI job. We just have to make sure we name the builders differently, otherwise they combine together in the CDash database. We shouldn't need to remove the build directory, but will need to explicitly state OpenMP should be enabled on the second invocation.

I will update the MR to include this.

robertmaynard: I have updated the MR to have CDash submit both for OpenMP and No-OpenMP.

robertmaynard:
@salasoom Is there anything else I need to revise?

travis.sh
salasoom: I realized some machines we get on travis do not support openmp, which is why I wasn't seeing those tests in the output. However, I still support the `rm -rf ./*` because the new CMake code will upload the 1st build's test results (without openmp) twice if the 2nd build fails. Take a look on CDash at an OSX build on the general dashboard page: we expect 148 tests to be run when openmp is enabled, but see 130 for OSX builds with openmp.

salasoom: On second thought I think the right thing to do is to conditionally build the 2nd time only if openmp is available on the machine. If a CI machine is always expected to fail the openmp build, it will only clutter the dashboard.

robertmaynard: Do you know of a way to figure in the the travis script if the machine will support `OpenMP`?

salasoom: I don't see any direct easy way to check through the command line. I think we could make a short helper CMake script to run `find_package(OpenMP COMPONENTS C REQUIRED)`. The travis script could conditionally start a 2nd build, depending on that helper cmake code succeeded or failed.

PR #50, opened 2019-03-03
Fixes of linker errors and warnings when using Visual Studio 2017.

wmamrak:
Also some CMake improvements.


lindstro:
@wmamrak Thanks for your contribution.  We're busy working on the next zfp release but we'll take a closer look at your PR in the next few days.

egorpugin:
+1 for bad dllexport/dllimport system (this PR does not fix all issues).

Some general setup could look like
```
// first, protect our api def from user-defined thing (some build systems know this better for you!)
#ifndef ZFP_API
    // actually, it's not my business anymore
    #ifdef ZFP_STATIC
        #ifdef __cplusplus
            #define ZFP_API extern "C"
        #else
            #define ZFP_API extern
        #endif
    #else
        // following part is ok in current code
        #ifdef ZFP_SOURCE
            #ifdef __cplusplus
                #define ZFP_API extern "C" __declspec(dllexport)
            #else
                #define ZFP_API extern     __declspec(dllexport)
            #endif
        #else
            #ifdef __cplusplus
                #define ZFP_API extern "C" __declspec(dllimport)
            #else
                #define ZFP_API extern     __declspec(dllimport)
            #endif
        #endif
    #endif
#endif
```

lindstro:
@egorpugin Thank you for your contributions.  The section of code in question involves Windows specific constructs (__declspec) that are not part of the C language standard, although as best I can tell, your proposed changes do not seem directly related to that.  Can you please elaborate on what problem your proposed code addresses and how?  Is the idea for the build system to define ZFP_STATIC when a static library (not DLL) is being built?  If so, then I imagine additional CMake changes are needed.  This is probably obvious to Windows developers, but that's not our primary platform, so any help with this would be appreciated.

egorpugin:
> Is the idea for the build system to define ZFP_STATIC when a static library (not DLL) is being built? If so, then I imagine additional CMake changes are needed. 

Yes and yes.
Also my example does not cover `__attribute__ visibility` case for non windows platforms.

Good example can be found https://gcc.gnu.org/wiki/Visibility
What is right there is that `extern` should not be cared near `ZFP_API` thing.

It must be handled separately.

salasoom:
@wmamrak Your commits regarding compiling the CFP example, and fixing ZFP compile definitions were taken care of in the latest 0.5.5 release (master branch), and on develop. We cannot merge half of a pull request, so it may be easier to submit separate PRs in the future for separate issues.

https://github.com/LLNL/zfp/commit/e86c98856e4b17a2b6efb8ffca07f7e61dc77ab3
https://github.com/LLNL/zfp/commit/e8b863418c14750f8e8558e1395cb452de00526d

We have also now updated the default branch to `develop`. Please modify the PR (to rebase off of develop), or submit a new one from the develop branch (which should now be the default chosen branch).

wmamrak:
I will make a new PR.

Issue #54, opened 2019-03-29
cfp library not part of install package


alichnewsky:
pull request upcoming.

lindstro:
@alichnewsky Thanks for your PR.  This issue is now addressed on the develop branch and will be included in the next release.

PR #56, opened 2019-03-29
package CFP if built

alichnewsky:
addresses issue #54
I have tested build/install on macOS and Linux. 

now rebased onto develop branch


salasoom:
(reapplied this commit via cherry-pick into develop for linear history)

PR #61, opened 2019-05-30
CUDA optimization: unrolling optimization gather_partial functions

maddyscientist:
Optimization of CUDA encoder.  The `gather_partial` functions use a runtime variable for their for loop trip count.  This means the compiler cannot unroll the loops and this results in register spillage since the compiler cannot index into registers.

This optimization leads to a 1.18x speedup on the benchmark when running on GP100
`zfp -d -3 384 384 256 -r 16 -x cuda -i ~/zfp/viscosity.raw`

### DIFF
--- a/src/cuda_zfp/encode1.cuh
+++ b/src/cuda_zfp/encode1.cuh
@@ -17,8 +17,8 @@ __device__ __host__ inline
 void gather_partial1(Scalar* q, const Scalar* p, int nx, int sx)
 {
   uint x;
-  for (x = 0; x < 4; x++, p += sx)
-    if (x < nx) q[x] = *p;
+  for (x = 0; x < 4; x++)
+    if (x < nx) q[x] = p[x * sx];
   pad_block(q, nx, 1);
 }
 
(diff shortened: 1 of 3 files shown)


lindstro:
@maddyscientist Thanks for this PR.  I am not convinced that this optimization does the right thing, however.  In `gather_partial2`, p is incremented by nx * sx in the inner loop without the optimization, which is then undone in the outer loop by subtracting nx * sx.  With the optimization, p is incremented by 4 * sx.  So in the outer loop, I would expect a subtraction of 4 * sx.

I am also concerned about p being made to potentially point to unallocated memory.  I'm not sure what CUDA says about this, but that's undefined behavior in C and C++, even if the pointer is not dereferenced.

maddyscientist:
Thanks for the review @lindstro.  I'm traveling in China at the moment so I might not get a chance to address your comments until I return.

One question: is there a suitable CUDA unit test that I can use that will test the `gather_partial` function?

lindstro:
@maddyscientist We do not have a unit test for `gather_partial` alone.  The `testZfpCuda*` tests should all indirectly call `gather_partial`, but using only a limited set of array dimensions that do a poor job of exercising this function.  We're working on adding more tests that should test virtually all possible combinations of partial block dimensions and that would expose any issues with its implementation.  This will require a fairly significant rewrite of our checksumming capability, as this will result in hundreds or even thousands of new checksums that are currently entered by hand.

Do the `testZfpCuda*` tests pass for you with your optimizations?  That would indicate a potential issue with the tests themselves.

maddyscientist:
@lindstro I have fixed the issues you highlighted above, and I can confirm that all the tests pass now.

PR #63, opened 2019-07-04
Fixes of linker errors and warnings when using Visual Studio 2017/2019.

wmamrak:
Related to PR #50

### DIFF
--- a/src/cuda_zfp/encode.cuh
+++ b/src/cuda_zfp/encode.cuh
@@ -74,7 +74,7 @@ fwd_lift(Int* p)
   Int x = *p; p += s;
   Int y = *p; p += s;
   Int z = *p; p += s;
-  Int w = *p;
+  Int w = *p; p += s;
 
   // default, non-orthogonal transform (preferred due to speed and quality)
   //        ( 4  4  4  4) (x)
@@ -87,7 +87,7 @@ fwd_lift(Int* p)
   w += y; w >>= 1; y -= w;
   w += y >> 1; y -= w >> 1;
 
-          *p = w;
+  p -= s; *p = w;
   p -= s; *p = z;
   p -= s; *p = y;
   p -= s; *p = x;
--- a/src/cuda_zfp/encode.cuh
+++ b/src/cuda_zfp/encode.cuh
@@ -74,7 +74,7 @@ fwd_lift(Int* p)
   Int x = *p; p += s;
   Int y = *p; p += s;
   Int z = *p; p += s;
-  Int w = *p; p += s;
+  Int w = *p;
 
   // default, non-orthogonal transform (preferred due to speed and quality)
   //        ( 4  4  4  4) (x)
@@ -87,7 +87,7 @@ fwd_lift(Int* p)
   w += y; w >>= 1; y -= w;
   w += y >> 1; y -= w >> 1;
 
-  p -= s; *p = w;
+          *p = w;
   p -= s; *p = z;
   p -= s; *p = y;
   p -= s; *p = x;
--- a/cfp/src/CMakeLists.txt
+++ b/cfp/src/CMakeLists.txt
@@ -8,6 +8,7 @@ list(APPEND cfp_private_defs ${zfp_compressed_array_defs})
 
 if(WIN32 AND BUILD_SHARED_LIBS)
   # define ZFP_SOURCE when compiling libcfp to export symbols to Windows DLL
+  list(APPEND cfp_public_defs ZFP_SHARED_LIBS)
   list(APPEND cfp_private_defs ZFP_SOURCE)
 endif()
 
--- a/cfp/src/CMakeLists.txt
+++ b/cfp/src/CMakeLists.txt
@@ -6,7 +6,7 @@ endif()
 
 list(APPEND cfp_private_defs ${zfp_compressed_array_defs})
 
-if(WIN32)
+if(WIN32 AND BUILD_SHARED_LIBS)
   # define ZFP_SOURCE when compiling libcfp to export symbols to Windows DLL
   list(APPEND cfp_private_defs ZFP_SOURCE)
 endif()
--- a/cfp/src/cfparrays.cpp
+++ b/cfp/src/cfparrays.cpp
@@ -7,7 +7,7 @@
 #include "cfparray3f.cpp"
 #include "cfparray3d.cpp"
 
-export_ const cfp_api CFP_NAMESPACE = {
+const cfp_api CFP_NAMESPACE = {
   // array1f
   {
     cfp_array1f_ctor_default,
--- a/include/zfp/system.h
+++ b/include/zfp/system.h
@@ -8,7 +8,7 @@
 #endif
 
 /* macros for exporting and importing symbols */
-#if defined _MSC_VER && defined ZFP_SOURCE
+#if defined _MSC_VER && defined ZFP_SHARED_LIBS
   #define export_ __declspec(dllexport)
   /* export (import) symbols when ZFP_SOURCE is (is not) defined */
   #ifdef ZFP_SOURCE
@@ -24,7 +24,7 @@
       #define extern_ extern     __declspec(dllimport)
     #endif
   #endif
-#else /* !_MSC_VER || !ZFP_SOURCE */
+#else /* !_MSC_VER || !ZFP_SHARED_LIBS */
   #define export_
   #ifdef __cplusplus
     #define extern_ extern "C"
--- a/include/zfp/system.h
+++ b/include/zfp/system.h
@@ -8,7 +8,7 @@
 #endif
 
 /* macros for exporting and importing symbols */
-#ifdef _MSC_VER
+#if defined _MSC_VER && defined ZFP_SOURCE
   #define export_ __declspec(dllexport)
   /* export (import) symbols when ZFP_SOURCE is (is not) defined */
   #ifdef ZFP_SOURCE
@@ -24,7 +24,7 @@
       #define extern_ extern     __declspec(dllimport)
     #endif
   #endif
-#else /* !_MSC_VER */
+#else /* !_MSC_VER || !ZFP_SOURCE */
   #define export_
   #ifdef __cplusplus
     #define extern_ extern "C"
--- a/include/zfp/system.h
+++ b/include/zfp/system.h
@@ -9,7 +9,6 @@
 
 /* macros for exporting and importing symbols */
 #if defined _MSC_VER && defined ZFP_SHARED_LIBS
-  #define export_ __declspec(dllexport)
   /* export (import) symbols when ZFP_SOURCE is (is not) defined */
   #ifdef ZFP_SOURCE
     #ifdef __cplusplus
@@ -25,7 +24,6 @@
     #endif
   #endif
 #else /* !_MSC_VER || !ZFP_SHARED_LIBS */
-  #define export_
   #ifdef __cplusplus
     #define extern_ extern "C"
   #else
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,7 +36,7 @@ if(HAVE_LIBM_MATH)
   target_link_libraries(zfp PRIVATE m)
 endif()
 
-if(WIN32)
+if(WIN32 AND BUILD_SHARED_LIBS)
   # Define ZFP_SOURCE when compiling libzfp to export symbols to Windows DLL
   list(APPEND zfp_private_defs ZFP_SOURCE)
 endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -38,6 +38,7 @@ endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)
   # Define ZFP_SOURCE when compiling libzfp to export symbols to Windows DLL
+  list(APPEND zfp_public_defs ZFP_SHARED_LIBS)
   list(APPEND zfp_private_defs ZFP_SOURCE)
 endif()
 
--- a/src/bitstream.c
+++ b/src/bitstream.c
@@ -1,4 +1,4 @@
 #include "bitstream.h"
 #include "inline/bitstream.c"
 
-export_ const size_t stream_word_bits = wsize;
+const size_t stream_word_bits = wsize;
--- a/src/cuda_zfp/shared.h
+++ b/src/cuda_zfp/shared.h
@@ -222,7 +222,7 @@ inv_lift(Int* p)
 	x = *p; p += s;
 	y = *p; p += s;
 	z = *p; p += s;
-	w = *p; p += s;
+	w = *p;
 
 	/*
 	** non-orthogonal transform
@@ -237,7 +237,7 @@ inv_lift(Int* p)
 	y += z; z <<= 1; z -= y;
 	w += x; x <<= 1; x -= w;
 
-	p -= s; *p = w;
+	        *p = w;
 	p -= s; *p = z;
 	p -= s; *p = y;
 	p -= s; *p = x;
--- a/src/cuda_zfp/shared.h
+++ b/src/cuda_zfp/shared.h
@@ -222,7 +222,7 @@ inv_lift(Int* p)
 	x = *p; p += s;
 	y = *p; p += s;
 	z = *p; p += s;
-	w = *p;
+	w = *p; p += s;
 
 	/*
 	** non-orthogonal transform
@@ -237,7 +237,7 @@ inv_lift(Int* p)
 	y += z; z <<= 1; z -= y;
 	w += x; x <<= 1; x -= w;
 
-	        *p = w;
+	p -= s; *p = w;
 	p -= s; *p = z;
 	p -= s; *p = y;
 	p -= s; *p = x;
--- a/src/template/decode.c
+++ b/src/template/decode.c
@@ -12,7 +12,7 @@ _t1(inv_lift, Int)(Int* p, uint s)
   x = *p; p += s;
   y = *p; p += s;
   z = *p; p += s;
-  w = *p;
+  w = *p; p += s;
 
   /*
   ** non-orthogonal transform
@@ -27,7 +27,7 @@ _t1(inv_lift, Int)(Int* p, uint s)
   y += z; z <<= 1; z -= y;
   w += x; x <<= 1; x -= w;
 
-          *p = w;
+  p -= s; *p = w;
   p -= s; *p = z;
   p -= s; *p = y;
   p -= s; *p = x;
--- a/src/template/decode.c
+++ b/src/template/decode.c
@@ -12,7 +12,7 @@ _t1(inv_lift, Int)(Int* p, uint s)
   x = *p; p += s;
   y = *p; p += s;
   z = *p; p += s;
-  w = *p; p += s;
+  w = *p;
 
   /*
   ** non-orthogonal transform
@@ -27,7 +27,7 @@ _t1(inv_lift, Int)(Int* p, uint s)
   y += z; z <<= 1; z -= y;
   w += x; x <<= 1; x -= w;
 
-  p -= s; *p = w;
+          *p = w;
   p -= s; *p = z;
   p -= s; *p = y;
   p -= s; *p = x;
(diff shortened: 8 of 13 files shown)


lindstro:
@wmamrak Thanks for your contribution.  Many tests do not pass on AppVeyor.  Some failures are related to CDash issues, but there also seem to be some fundamental problems with this PR.  Would you have some time to look into this and resubmit a passing PR?

salasoom:
@wmamrak Build errors can be viewed here https://open.cdash.org/viewBuildError.php?buildid=5990665

It appears that variables marked as `extern` are not being linked properly.

src/cuda_zfp/encode.cuh
```diff
@@ -74,7 +74,7 @@ fwd_lift(Int* p)
   Int x = *p; p += s;
   Int y = *p; p += s;
   Int z = *p; p += s;
-  Int w = *p; p += s;
+  Int w = *p;
```
lindstro: I would expect any decent compiler to optimize away `p += s; p -= s;`.  This change reduces code readability, in my opinion, since it suggests that w is treated separately from x, y, and z.  Do you have any evidence that this optimization is useful?

wmamrak: Do you have any evidence that the compiler optimizes this away? 
In my opinion this change improves code readability as there are less operations to be analyzed (and no unnecessary ones, which means the code has been more carefully thought out). As a side note, why didn't you complain about code readability here https://github.com/LLNL/zfp/blob/develop/src/cuda_zfp/decode.cuh#L126 ? :-)

mclarsen: Its fairly easy to check what the compiler is optimizing.

Before: https://godbolt.org/z/c_6gvx
After: https://godbolt.org/z/i0caJO

You can change the compiler and flags to see the effects.

lindstro: > Do you have any evidence that the compiler optimizes this away?

With all due respect, the burden is not on us to verify the utility of your PR.  That said, gcc 4.8.5 does indeed optimize the code away.  I would be happy to reconsider if you can find a relevant compiler that does not perform this optimization.

> In my opinion this change improves code readability as there are less operations to be analyzed (and no unnecessary ones, which means the code has been more carefully thought out).

I take it then that you would rewrite this loop

```
  for (i = 0; i < 10; i++)
      a[i] += b[i];
```

as the more carefully thought out version

```
  for (i = 0; i < 9; i++)
      a[i] += b[i];
  a[9] += b[9];
```

since the latter has no unnecessary `i++`? ;-)  I think we'll just have to agree that readability is a matter of taste.

> As a side note, why didn't you complain about code readability here https://github.com/LLNL/zfp/blob/develop/src/cuda_zfp/decode.cuh#L126 ? :-)

Because I wrote the code? ;-)  I fully agree that this section of code is extremely obfuscated.  It's also the innermost loop and biggest performance bottleneck in all of zfp, so here readability takes a back seat.

wmamrak: The entire discussion starts to look more and more like a farce to me.
First you argument, that this minor change reduces readability, then you admit, that readability is a matter of taste. Reducing readability without reducing readability. 
You then ask about the utility of this change. Here it is: you don't have to rely on compiler optimizations to generate optimal code. This sounds very reasonable. And because of that it is not me who should prove anything, it is up to you to prove that this change is unnecessary, i.e. that compilers do indeed remove the redundant operations away. If I made a semantic change, then yes, it would be on me to show the behaviour remained unchanged. But it is not the case. 

I will leave your exagarrated code sample without a comment as it is exagarrated, oversimplified, and shows a lack of respect.

To me this code change is: minor, prosaic, simplifying and, last but not least, pragmatic. Also, my guess is, that this code simplification has been overlooked in the original code. I believe, that in this paragraph alone I gave enough arguments behind it. If you don't agree with them, I am fine with it, I will revert the change, and please review the original PR, which is of much greater importance.

> > As a side note, why didn't you complain about code readability here https://github.com/LLNL/zfp/blob/develop/src/cuda_zfp/decode.cuh#L126 ? :-)
> 
> Because I wrote the code? ;-) I fully agree that this section of code is extremely obfuscated. It's also the innermost loop and biggest performance bottleneck in all of zfp, so here readability takes a back seat.

Does this obfuscation improve performance? Because I had to unobfuscate it as direct translation to HLSL resulted in incorrect behaviour.

lindstro: I'm sorry you feel this way.  I threw in a couple of winks to indicate that I was only half joking and I apologize if my response came across as sarcastic.

Just to be clear, I said "This change reduces code readability, **in my opinion**..."  I wasn't trying to argue that code readability is absolute, or we'd all be using the same coding style. To me, your change to our code reduces its readability with no apparent benefit since compilers tend to optimize away such code, just like they can sometimes optimize away the redundant i++ at the end of a loop.

I do not buy the argument that programmers should micro optimize code to "help" the compiler just in case it cannot perform trivial optimizations.  If the compiler cannot make this particular optimization, chances are that this line of code isn't your biggest problem...

The loop example was a bit tongue-in-cheek, but the point was that we often write "redundant" code (the i++, in this case) for the sake of readability/simplicity/maintainability.  Your suggested optimization is equivalent to removing the final i++ in an unrolled loop, which we apparently both agree would be a silly and unnecessary optimization.

I'm honestly puzzled that you believe that it is my responsibility to disprove that your unsolicited contribution has no effect on the dozens of compilers we support rather than you providing one example where it is beneficial.  Toward disproving utility, how am I to know I didn't miss a compiler?

If you want your submission to be adopted in our code base, then it falls on you to give us reasonable evidence that it is beneficial, e.g., by documenting any performance improvements (see, e.g., #61 and #67).

Regarding obfuscated code in the encoder/decoder, yes, this admittedly atrocious-looking piece of code did run measurably faster than some alternative implementations we considered.  One problem in C is that you cannot break out of more than a single loop level, which often necessitates redundant loop conditionals, extra branches, or goto statements, none of which are particularly attractive alternatives.

Finally, it is easier for us to manage pull requests if they're isolated and address individual issues.  Please don't bundle micro optimizations with unrelated Windows linker issues.  I will get to the Windows issues in time but am currently swamped with other tasks.  Would you mind summarizing what problem your Windows PR solves and why these linker issues do not seem to manifest themselves when testing on AppVeyor?

lindstro: > Its fairly easy to check what the compiler is optimizing.
> 
> Before: https://godbolt.org/z/c_6gvx
> After: https://godbolt.org/z/i0caJO
> 
> You can change the compiler and flags to see the effects.

@mclarsen Interesting.  So the proposed optimization had no effect in this case.  Having disassembled the low-level zfp code many times myself, this is also my experience.  This particular build does not exploit SSE registers or IPO; in practice, the values being operated on are kept in registers and are not fetched from memory, as the compiler can statically determine that all addresses refer to a locally declared array.  So even `x = *p;` and all `p += s;` statements are entirely optimized away (at least by gcc).

I wasn't aware of Compiler Explorer.  Neat tool.  Are there web hooks for GitHub?  How did you point it to the zfp repo?  Or did you just cut-and-paste this function?

mclarsen: @lindstro I don't believe there are any web hooks to GitHub. This tool is more for exploring small amounts of code, and I just copied the function over. I only looked at the gcc 7 versions of the function and I am not sure what ISA its targeting. I really just wanted to make you aware of this tool, which I think is great.

wmamrak: I took no offense, but If someone would make an unsolicited contribution to my code, then even if this contribution was a micro optimization, the one that is usually performed by a compiler, then assuming it is correct, I would welcome it happily, as opposed to neglecting its usefulness by using subjective measures of code readability. And I do not agree this is similar to redundant i++. I will exagarrate a bit now, but original code is more like:
```
for (i = 0; i < 10000; i++) {
    i += 2;
    i -= 2;
    a[i] = b[i];
}
```
This is a typical, unnecessary redundancy which in most cases reduces legibility and is universally treated as an error, even though the compiler should optimize the redundant code away. I believe you agree, that the above redundancy has no relation to your redundant increment example.

>I'm honestly puzzled that you believe that it is my responsibility to disprove that your unsolicited contribution has no effect on the dozens of compilers we support rather than you providing one example where it is beneficial. Toward disproving utility, how am I to know I didn't miss a compiler?

Proposed change has objectively non-negative impact, whatever compiler is used. In contrary, not applying it may have negative impact.

>Finally, it is easier for us to manage pull requests if they're isolated and address individual issues.

I admit to a mistake.

wmamrak:
>Would you mind summarizing what problem your Windows PR solves?

Yes.
Assume we are compiling a shared zfp library. BUILD_SHARED_LIBS=1, hence ZFP_SOURCE is defined, thus symbols are exported by means of dllexport keyword.
Now assume there is another program we would like to link against our zfp's DLL. In this program ZFP_SOURCE is not defined, so we would expect, that symbols are imported this time. They are not, though, because ZFP_SOURCE condition in the outer #if in system.h (line no. 11) prevents it. dllimport keyword is not used which leads to linker errors.
To remove the ZFP_SOURCE condition is not an option, so the proposed solution is to define *public* preprocessor definition (ZFP_SHARED_LIBS) which propagates from zfp imported target to all the targets linked against it, and which helps in resolving whether dllimport keyword should be used, or not.

>why these linker issues do not seem to manifest themselves when testing on AppVeyor?

The tests do not cover the cases of linking against DLLs.

P.S. I don't think ZFP_SOURCE is the most descriptive and adequate name.

src/cuda_zfp/encode.cuh
lindstro: I agree that your example exhibits unnecessary redundancy that ostensibly serves no purpose.  I also agree that this example of yours has no relation to eliminating the last `p += s;` in the for loop below, which is also why I think your example is a poor analogy.

Consider this piece of code:

```
for (i = 0; i < 4; i++) {
  a[i] = *p;
  p += s;
}
```

This unrolls to

```
  a[0] = *p; p += s;
  a[1] = *p; p += s;
  a[2] = *p; p += s;
  a[3] = *p; p += s;
```

Clearly the last `p += s;` is not necessary here--or in the original loop, which can be rewritten with some effort to omit this statement in a somewhat obfuscated way--unless of course this sequence of statements is implemented as a single loop instruction like `rep movs`, in which case the final `p += s;` is necessary and the "optimization" is actually harmful.

Non-essential changes can also have unintended consequences.  Who's to say that there isn't a zfp fork out there that has generalized zfp's encoder to operate on 8 values at a time and that depends on the fourth `p += s;` being executed?

My point is that all things being equal, unless there actually is a tangible benefit, then this code change reduces to one of style.  I suspect you would hesitate to adopt a style change (like my earlier contrived loop example) in your own code for the sake of style only--especially if you prefer your own style.

I prefer the `p += s;` at the end because it maintains symmetry with the other statements and I do not need to parse the code to figure out why this one line of code is different from the other lines.  Sure, this is subjective, but so is your opinion on the matter until you have demonstrated a clear benefit. :-)

lindstro:
@wmamrak Thanks for explaining the linker issue.  I'm still a little confused.  Why does line 11 (`#ifdef _MSC_VER`) prevent the dllimport?

I must confess to not understanding why Windows requires these non-standard language extensions that Linux and macOS do fine without.  I'm not sure about the origins of the ZFP_SOURCE macro, though I suspect that it came from some other PR or something that perhaps the Kitware folks contributed.  I'm willing to entertain other names (any suggestions?).  I believe this macro is intended to be defined only when compiling the zfp source code, not when including zfp headers in an application.

Clearly this must be something that is common when writing code for Windows.  Do you know of other libraries or documentation that illustrate this Windows-specific idiom?

src/cuda_zfp/encode.cuh
wmamrak: This example exhibits unnecessary redundancy that ostensibly serves no purpose *and* it has been directly taken from the original code. This time it is not an analogy, it is literally *the* actual implementation.
```
for (i = 0; i < 4; i++) {
  Int w = *p;
  p += s;
  transform();
  p -= s;
  *p = w;
}
```

>Non-essential changes can also have unintended consequences. Who's to say that there isn't a zfp fork out there that has generalized zfp's encoder to operate on 8 values at a time and that depends on the fourth p += s; being executed?

That is the point of forks - to differ in some way from the original code - isn't it? And this particular fork would differ to such an extent, that such a trivial change would have never confused its maintainer.

>My point is that all things being equal, unless there actually is a tangible benefit, then this code change reduces to one of style. I suspect you would hesitate to adopt a style change (like my earlier contrived loop example) in your own code for the sake of style only--especially if you prefer your own style.

I regret you did not state this in the very first post, i.e. that you do not accept this change because you adore the code in which redundant operations are optimized out by the compiler, that would save me quite some time! :-)
One last word about symmetry, which I adore myself - sometimes humans artificially create symmetries where there ain't any, and I believe this is the case here.

I will revert the last commit to save both of us from further philosophical considerations. I would never have expected such a minor change can trigger such a discussion.

wmamrak:
> @wmamrak Thanks for explaining the linker issue. I'm still a little confused. Why does line 11 (`#ifdef _MSC_VER`) prevent the dllimport?

Ohh, I was referring to an incorrect base file. Ignore my previous post, please.

Currently, when compiling a static library, there are a few compiler warnings of the form:
>warning C4273: 'zfp_codec_version': inconsistent dll linkage

The reason is these symbols are using __declspec(dllimport) keyword.

When linking against a static library, there are linker warnings of the form:
>warning LNK4217: symbol '_stream_word_bits' defined in 'zfp.lib(bitstream.obj)' is imported by 'main_compress.obj' in function '_main'
warning LNK4286: symbol '_stream_word_bits' defined in 'zfp.lib(bitstream.obj)' is imported by 'zfp.lib(zfp.obj)'

These follow from the compiler errors mentioned previously.

The introduction of a public ZFP_SHARED_LIBS preprocessor definition and the new term in the condition `#if defined _MSC_VER && defined ZFP_SHARED_LIBS` solves the above issues because the correct export keywords are used (i.e. none).

>Clearly this must be something that is common when writing code for Windows. Do you know of other libraries or documentation that illustrate this Windows-specific idiom?

It is. See https://docs.microsoft.com/en-us/cpp/build/importing-into-an-application-using-declspec-dllimport?view=vs-2019
But this idiom is not enough when mixing static and shared libraries, and that is what public ZFP_SHARED_LIBS macro resolves.

wmamrak:
Also, the export_ macro seems to not be needed. It is empty on non msvc, and is used with definitions of global variables whose definitions use extern_ macro, which is enough.

wmamrak:
Please forget everything you "learned" from this MSDN link :-)
Here is a better one: https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/
And its concise version: https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html

zfp sets WINDOWS_EXPORT_ALL_SYMBOLS to true, so all the functions are automatically exported by means of .def file, so there is no need to annotate them with __declspec(dllexport).
Unfortunately exporting global variables, like zfp_version_string, is much trickier, and CMake does not even attempt to do this, hence we need to mark them manually with __declspec(dllexport) when compiling zfp dll, and with __declspec(dllimport) when compiling against the code in the zfp dll.
See also the very first post here: https://github.com/libressl-portable/portable/issues/223

I believe now all your doubts disappeared.

>The changes you made remove _declspec(dllexport) from variable definitions. Why? The extern macro is not applied to those definitions.

In my version of the code all five global variables' declarations are annotated with extern macro. :-)

>Is ZFP_SHARED_LIBS defined when zfp is built or when it is used?

In both cases - it is a public preprocessor definition that is defined for zfp imported target and propagates to all targets that link against it. It is in contrary to ZFP_SOURCE preprocessor definition which is private, i.e. it is defined only during zfp's build.

>What do you mean by "mixing static and shared libraries" in this case?

That was a really poor wording of mine :-) What I meant was, that this dllexport/dllimport idiom is fine for building and linking against shared libraries, but needs to be extended to support static ones.

>Sorry for all these questions, but I frankly do not understand the purpose of all this extra work and why the Windows compiler cannot resolve these issues itself.

I never investigated that myself.

lindstro:
Ah, I do remember now this CMake feature, which doesn't export data symbols, which is why we added the `export_` macro.

The new documentation you pointed me at makes more sense, but I'm still a bit confused:

"For global data symbols, __declspec(dllimport) must still be used when compiling against the code in the DLL. The symbol is exported from the DLL correctly and automatically, but the compiler needs to know that it is being imported from a DLL at compile time."

Later on:

"Static const members: These are not exported by the new feature.  In some cases to support older compilers, these are put in .cxx files and compiled into the library and are not just inlined in the .h files.  If you have this case, you will need to use dllimport and dllexport markup and not just dllimport mark up as the auto export will not work."

This suggests that dllexport is still needed.  Perhaps this comment refers only to static const class members and not to global const data (though it's unclear why there would be a difference).

Anyway, from this reading, all that is needed is the addition of dllimport for all data declarations by users of zfp when zfp has been built as a DLL.  Do you agree?

So if I understand correctly, ZFP_SOURCE will be defined when zfp is built (whether as a static or shared library).  This will suppress any dllimport attributes.  When ZFP_SHARED_LIBS is defined and ZFP_SOURCE isn't, dllimport will be added to public data declarations in zfp header files.  Did I get this right?  Is dllexport not needed?

> In my version of the code all five global variables' declarations are annotated with extern macro. :-)

`extern` is a keyword.  Do you mean `extern_`?  I would strongly suggest renaming this macro if you're using it for this purpose.  `extern_` could be confused for `extern`.  `extern_` really should mean "export this symbol" in this context, while `extern` roughly means "import this symbol" (i.e., this symbol is defined elsewhere).  That's why we named this macro `export_`. :-)

wmamrak:
> Ah, I do remember now this CMake feature, which doesn't export data symbols, which is why we added the `export_` macro.

The export_ macro was used incorrectly, i.e. it annotated global data definitions. Declarations were annotated by extern_ macro. I have fixed that minor bug while removing the export_ macro.

> This suggests that dllexport is still needed. Perhaps this comment refers only to static const class members and not to global const data (though it's unclear why there would be a difference).

Yes, it says `Static const *members*`, although after reading the documentation, initially I though all global symbols will be automatically exported. The third link I provided explains this is not the case very clearly:

>CMake EXPORT_ALL only exports functions, not data (e.g. global variables). Exporting data from a Windows DLL is complicated, and CMake is right not to attempt it. You can export data symbols in a DEF file, but it probably won't end well.
>
>IMHO, the only sane way to handle exported data on Windows is to annotate the declarations in the header file. Use __declspec(dllexport) when building the DLL, __declspec(dllimport) when using the DLL, and no annotation when building or using a static library.


> Anyway, from this reading, all that is needed is the addition of dllimport for all data declarations by users of zfp when zfp has been built as a DLL. Do you agree?

Yes.

> So if I understand correctly, ZFP_SOURCE will be defined when zfp is built (whether as a static or shared library). This will suppress any dllimport attributes. 

Yes and yes. And also I would consider renaming it to something more meaningful.

> When ZFP_SHARED_LIBS is defined and ZFP_SOURCE isn't, dllimport will be added to public data declarations in zfp header files. Did I get this right? 

Correct.

> Is dllexport not needed?

Yes, it is not needed at this stage anymore, anywhere.

> > In my version of the code all five global variables' declarations are annotated with extern macro. :-)
> 
> `extern` is a keyword. Do you mean `extern_`? I would strongly suggest renaming this macro if you're using it for this purpose. `extern_` could be confused for `extern`. `extern_` really should mean "export this symbol" in this context, while `extern` roughly means "import this symbol" (i.e., this symbol is defined elsewhere). That's why we named this macro `export_`. :-)

Yes, I meant `extern_` macro (I used `macro` in that sentence explicitly to stress that fact out :-)) as opposed to `extern` storage class specifier. I thought you meant the former as well, and then I got confused, because you seemed to be working on a different code base. :-)
I don't know who came up with that macro's name, but I would suggest to rename it as well. I would have done it already, but I didn't want to introduce more changes before being sure you are ok with the previous ones.

To sum up, to my best knowledge, and according to my local tests, after applying the changes from this PR, compilation of and linkage against both static and shared zfp libraries under Windows succeeds without errors and warnings. What is left is possible renaming of macros, which I leave up to you :-)

PR #65, opened 2019-07-17
Let the compiler inline the permutation lookup table into registers

maddyscientist:
This pull request is an optimization for the CUDA backend, for both encoding and decoding.
* Previously, the permutation lookup table was stored in `__constant__` memory, this means the compiler cannot know the contents of the look-up table and so is forced to index from local memory rather than registers.
* Instead, we store the lookup table directly in registers by simply marking up the lookup table with the `__device__` qualifier.  Since the array is marked as `const` the compiler is able to evaluate the look-up table values at compile time and as a result all register spills / local memory associated with this are removed.
* One nice side effect is that the code is simplified since all use of `__constant__` memory is removed.

This optimization leads to a 1.25x speedup in the following encoding benchmark (on top of the 1.18x speedup already from #61 - though this pull is completely independent of that). 
`zfp -d -3 384 384 256 -r 16 -x cuda -i ~/zfp/viscosity.raw`

All tests are confirmed to run after this change.


lindstro:
@maddyscientist Many thanks for this updated PR.  The performance numbers are impressive!  At first glance, I see no reason not to merge this PR, but let me consult with @mclarsen first.  We'll get back to you shortly.

maddyscientist:
No problem @lindstro.  Let me just clarify that the speedup numbers are for the encode kernel itself only, and do not include the CPU-GPU transfer times that are also done in the above referenced benchmark.

PR #68, opened 2019-08-14
Add short note and link for h5z-zfp

markcmiller86:
Wasn't sure of best place to put. Added to end of section on the zfp file compression command. I adjusted the title of the section slightly so that it was a little more inclusive.

### DIFF
--- a/docs/source/zfpcmd.rst
+++ b/docs/source/zfpcmd.rst
@@ -5,6 +5,11 @@
 File Compression
 ================
 
+This section describes a simple, file-based |zfp| compression tool that is
+part of the |zfp| distribution named |zfpcmd|. If you are interested in
+other, third-party, file-based compression options, please be sure to read
+about those mentioned in the :ref:`Overview <overview>` section.
+
 The |zfpcmd| executable in the :file:`bin` directory is primarily
 intended for evaluating the rate-distortion (compression ratio and quality)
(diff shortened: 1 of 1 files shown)


lindstro:
@markcmiller86 I agree that we should advertise H5Z-ZFP, but this doesn't seem the right place for it as the intent of this section is to document the zfp command-line utility.  H5Z-ZFP and other plugins (for ADIOS, Intel IPP, TTK, ...) are discussed in the [overview](https://zfp.readthedocs.io/en/release0.5.5/overview.html) section.  Admittedly we could do a better job highlighting these plugins elsewhere in the documentation.

I don't think we want to endorse only HDF5, so perhaps the best solution is to add a reference from the File Compressor section back to the overview section and mention that there is zfp support for doing file I/O through other I/O libraries.  Does that make sense?

markcmiller86:
> I don't think we want to endorse only HDF5, so perhaps the best solution is to add a reference from the File Compressor section back to the overview section and mention that there is zfp support for doing file I/O through other I/O libraries. Does that make sense?

Sure. That makes perfect sense. I wasn't aware of those external references anywhere in these docs.

markcmiller86:
@lindstro Sorry for the delay. It wasn't clear to me if you wanted me to make the proposed change or someone else. I fiddled with the text a bit as you suggested putting a mention at top and bottom of the section.

PR #72, opened 2019-09-04
Added necessary changes to use ZFP in Kripke and RAJA

tristanvdb:
* Changes to cmake for integration with Kripke/RAJA
 * Made zfp::array1::pointer constructor public to use it in Kripke
 * Made ctor for zfp::arrayN::pointer public and set reference_t to zfp::arrayN::reference.
 * Added zfp::array::pointer::dims, zfp::array::pointer::size(), zfp::array::pointer::pointer(std::nullptr_t)

### DIFF
--- a/array/zfp/pointer1.h
+++ b/array/zfp/pointer1.h
@@ -2,6 +2,10 @@
 class pointer {
 public:
   pointer() : ref(0, 0) {}
+#if  __cplusplus >= 201103L
+  pointer(std::nullptr_t n) : ref(0, 0) {}
+#endif
+
   pointer operator=(const pointer& p) { ref.array = p.ref.array; ref.i = p.ref.i; return *this; }
   reference operator*() const { return ref; }
   reference operator[](ptrdiff_t d) const { return *operator+(d); }
@@ -27,4 +31,7 @@ protected:
   void increment() { ref.i++; }
   void decrement() { ref.i--; }
   reference ref;
+
+public:
+  array1 const & data() const { return *(ref.array); }
 };
--- a/array/zfp/pointer2.h
+++ b/array/zfp/pointer2.h
@@ -2,6 +2,10 @@
 class pointer {
 public:
   pointer() : ref(0, 0, 0) {}
+#if  __cplusplus >= 201103L
+  pointer(std::nullptr_t n) : ref(0, 0, 0) {}
+#endif
+
   pointer operator=(const pointer& p) { ref.array = p.ref.array; ref.i = p.ref.i; ref.j = p.ref.j; return *this; }
   reference operator*() const { return ref; }
   reference operator[](ptrdiff_t d) const { return *operator+(d); }
@@ -39,4 +43,7 @@ protected:
     }
   }
   reference ref;
+
+public:
+  array2 const & data() const { return *(ref.array); }
 };
--- a/array/zfp/pointer3.h
+++ b/array/zfp/pointer3.h
@@ -2,6 +2,10 @@
 class pointer {
 public:
   pointer() : ref(0, 0, 0, 0) {}
+#if  __cplusplus >= 201103L
+  pointer(std::nullptr_t n) : ref(0, 0, 0, 0) {}
+#endif
+
   pointer operator=(const pointer& p) { ref.array = p.ref.array; ref.i = p.ref.i; ref.j = p.ref.j; ref.k = p.ref.k; return *this; }
   reference operator*() const { return ref; }
   reference operator[](ptrdiff_t d) const { return *operator+(d); }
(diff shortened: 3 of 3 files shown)


array/zfp/pointer1.h
```diff
@@ -2,6 +2,7 @@
 class pointer {
 public:
   pointer() : ref(0, 0) {}
+  pointer(std::nullptr_t n) : ref(0, 0) {}
```
lindstro: std::nullptr_t is C++11 while zfp is C++98 compliant.  If this is absolutely necessary for RAJA, then appropriate ifdefs are needed to hide this ctor.

array/zfp/pointer1.h
```diff
@@ -21,10 +22,18 @@ class pointer {
   friend class array1;
   friend class reference;
   explicit pointer(reference r) : ref(r) {}
+public:
```
lindstro: Why should this ctor be made public?  It is available from the array being passed in, i.e., &array[i].

array/zfp/pointer1.h
```diff
@@ -21,10 +22,18 @@ class pointer {
...
   explicit pointer(array1* array, uint i) : ref(array, i) {}
+protected:
   ptrdiff_t index() const { return ref.i; }
   void set(ptrdiff_t index) { ref.i = index; }
   void increment() { ref.i++; }
   void decrement() { ref.i--; }
   reference ref;
+
+public:
```
lindstro: More C++11 stuff.  Does RAJA need the type alias?  I want to make sure dims and size() are needed.  It is somewhat unnatural to ask for a pointer's size and have it return the size of the array into which is points.  There's no equivalent operator for native pointers.

array/zfp/pointer1.h
tristanvdb: It is used call `std::vector<ptr_t>::resize` in the same way whether `ptr_t = double *` or `ptr_t = zfp::array::pointer`.

I will add the proper `#ifdef`

array/zfp/pointer1.h
tristanvdb: In Kripke, I use `baseptr = typename type::pointer(store, 0);`. I can replace it by `baseptr = &(store[0]);`.

array/zfp/pointer1.h
tristanvdb: I removed reference_t and dims.

However, I added data() which returns the underlying array. It is needed to support View1D (I need to know the size of the data contained by the underlying ZFP array).

lindstro: I'm still not convinced that a `data()` method is needed.  You should think of `zfp::array<double>::pointer` as a proxy for `double*`, and these types should in principle have identical semantics.  When initializing `double* ptr = &my_stl_vector[0];`, there's no way to query how much storage was allocated for (or any other information about) the vector via `ptr` alone (i.e., this information is not implicit).  Why is such a method needed for zfp arrays but not for uncompressed arrays (e.g., STL vectors) in the context of Kripke/Raja?

I'm not trying to make life more difficult for you, but I'm hesitant to add public API functions (plus documentation and tests) that seem nonessential and that may not even be implementable, e.g., if the `pointer` implementation one day changes in a way that precludes access to the underlying array.  I suspect there's a cleaner way of supporting what you want to do.  If you want zfp's pointer to behave more like a view, then I would suggest either using zfp's views or deriving your own view from zfp's pointer.

array/zfp/pointer1.h
lindstro: zfp's pointers are null initialized by default, so `pointer()` and `pointer(nullptr)` will achieve the same.  Is this ctor a (Kripke/Raja) requirement for a template interface?

If needed, I would still leave the unused parameter name off to avoid potential compiler warnings.

tristanvdb: It is needed to match the initialization of a `double *` to nullptr (minimizes template specialization). I removed the name.

array/zfp/pointer1.h
tristanvdb: I only need this method in the short term. I believe I will be able to obtain that information directly in RAJA::View.

tristanvdb:
Given the extra method on pointer which should not make it to master and the fact that I started working on the ARC branch, I will close this pull request until further notice.

lindstro:
I'd be happy to discuss this further and make sure that we both agree on what changes are needed, if any.  I'm hoping to consolidate some redundant zfp/ARC code.  For instance, make `pointer` templated on the surrounding container type (e.g., fixed- or variable-rate array).  You'll notice that `pointer` and `vpointer` are identical except for the reference to the underlying array.  The same applies to references and iterators.  I would hold off on pull requests until this has been ironed out.  Let's talk offline about any changes needed for Kripke/RAJA.

PR #83, opened 2019-11-24
hotfix: Fix wrong memset params order

mohamed:
The init value and length parameters of memset were swapped.
This commit fixes that.

Signed-off-by: Mohamed A. Bamakhrama <mohamed@alumni.tum.de>


lindstro:
@mohamed Thanks for finding this bug and submitting a PR.

Issue #64, opened 2019-07-09
Provide pip and/or conda installable Python package


impredicative:
It would really help to have a `pip` installable Python 3.7 compatible package published on https://pypi.org/ . This could then be updated whenever there is a new release of zfp. It will lower the barrier to using zfp. Thanks.

lindstro:
Agreed, but unfortunately our Python developer has moved on to another project.  Any assistance with this would be appreciated.  In case it's helpful, I do have a pypi account.

impredicative:
In that case, do you have a FTE opening for a Python developer?

lindstro:
I'm afraid I don't have sufficient funding for an FTE, nor do I have enough work to keep a Python developer busy long term.

jayvdb:
Dummies guide to packaging is to do `python setup.py test` , ensure that works, and then run PyPI `check-manifest` to ensure `MANIFEST.in` is complete, and then `python setup.py sdist` and use `twine` to upload it to PyPI.

Then get some python people to build it at home, fix any errors, etc, and then in the next release also do `python setup.py bdist_wheel` which is better for normal users.

kmpaul:
@lindstro Would there be interest in making a conda package for zfp/zfpy instead of pip?  Pip is nice for pure python packages, but it's not the best for python packages with special C/C++ dependencies.

If this would be of interest, I think I can direct some effort to this tomorrow.

kmpaul:
...Although, I agree that a wheel would be good, too.

impredicative:
@kmpaul Speaking as a user, yes, conda works fine.

lindstro:
@kmpaul Thanks for the offer!  Python is not my strength, and I'm not sure I'm qualified to answer your question.  Of course, any contributions that simplify life for our Python users are welcome.

I'm roping @GarrettDMorrison in on this, who is our resident Python expert.

kmpaul:
I definitely think that a conda forge package for zfp would be loved by many.  So, since I'm in a hackathon for the next day (or two), I'll go ahead and start making some progress on this.  If the result is something you like, we can transfer ownership.

kmpaul:
Of course!  And I'm happy to transfer ownership.  I understand that you all might be busy, though, so I happen to have some free cycles at the moment to do it.

lindstro:
Awesome!  Thank you so much for your contributions, Kevin.

kmpaul:
[...]

PR #90, opened 2020-03-10
Cython fix

da-wad:
Had to make these changes to get Cython 0.29.15 to accept building the Python bindings and tests, perhaps it was originally built with a less strict compiler? Hope this is useful to others, and may be a step towards Issue #64


lindstro:
@da-wad Thanks for this fix!

PR #100, opened 2020-06-13
[WIP] Use correct installation directory on windows with modern CMake

hmaarrfk:
If you are willing to update the minimum version of CMake to 3.12, we can work on cleaning this up quite a bit.

Otherwise, happy to keep this here as a WIP


hmaarrfk:
Follow progress here https://github.com/conda-forge/zfpy-feedstock/pull/6

lindstro:
@hmaarrfk Sorry for the delay in processing this PR.  Currently some tests fail on Travis, ostensibly due to CMake 3.9.4 not supporting FindPython (see https://open.cdash.org/build/6814168/configure).

Can we conditionally require CMake 3.12 when `BUILD_ZFPY` is on?  I know 3.12 is old by now, but I prefer to keep requirements to a minimum.

hmaarrfk:
I htink it is likely best to wait until FindPython is supported more fully by scikit-build, then update with that.

PR #106, opened 2020-09-17
replace bytes of compressed stream with uint8_t

halehawk:
I replaced type bytes with const uint8_t[::1], this way the compressed stream can be checked by ensure_ndarray of numcodecs, and also it prevents numpy array tobytes procedures. I tested with ensure_ndarray, and the default zfp tests.

### DIFF
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -260,7 +260,7 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
+    cdef const void* comp_data_pointer = <const void*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         <void *>comp_data_pointer,
@@ -329,12 +329,12 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-     const uint8_t[::1] compressed_data,
+    const uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
+    cdef const void* comp_data_pointer = <const void *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         <void *>comp_data_pointer,
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -261,7 +261,7 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = <char*>&compressed_data[0]
+    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
@@ -335,7 +335,7 @@ cpdef np.ndarray decompress_numpy(
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = <char *>&compressed_data[0]
+    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -264,7 +264,7 @@ cpdef np.ndarray _decompress(
     cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        <char*>comp_data_pointer,
+        <void *>comp_data_pointer,
         len(compressed_data)
     )
     cdef zfp_stream* stream = zfp_stream_open(bstream)
@@ -330,7 +330,7 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-     uint8_t[::1] compressed_data,
+     const uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
@@ -338,7 +338,7 @@ cpdef np.ndarray decompress_numpy(
     cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        <char *>comp_data_pointer,
+        <void *>comp_data_pointer,
         len(compressed_data)
     )
     cdef zfp_stream* stream = zfp_stream_open(bstream)
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -254,7 +254,6 @@ cpdef np.ndarray _decompress(
     double rate = -1,
     int precision = -1,
 ):
-    print(cython.__version__)
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
     if compressed_data is out:
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -254,7 +254,7 @@ cpdef np.ndarray _decompress(
     double rate = -1,
     int precision = -1,
 ):
-
+    print(cython.__version__)
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
     if compressed_data is out:
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -246,7 +246,7 @@ cdef _validate_4d_list(in_list, list_name):
         )
 
 cpdef np.ndarray _decompress(
-    const uint8_t[::1] compressed_data,
+    uint8_t[::1] compressed_data,
     zfp_type ztype,
     shape,
     out=None,
@@ -261,7 +261,7 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
+    cdef char* comp_data_pointer = <char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
@@ -330,12 +330,12 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-    const uint8_t[::1] compressed_data,
+    uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
+    cdef char* comp_data_pointer = <char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -5,6 +5,7 @@ import cython
 from libc.stdlib cimport malloc, free
 from cython cimport view
 from cpython cimport array
+from libc.stdint cimport uint8_t
 import array
 
 import itertools
@@ -245,7 +246,7 @@ cdef _validate_4d_list(in_list, list_name):
         )
 
 cpdef np.ndarray _decompress(
-    bytes compressed_data,
+    const uint8_t[::1] compressed_data,
     zfp_type ztype,
     shape,
     out=None,
@@ -260,7 +261,7 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef char* comp_data_pointer = <char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
@@ -329,12 +330,12 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-    bytes compressed_data,
+    const uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef char* comp_data_pointer = <char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
         comp_data_pointer,
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -246,7 +246,7 @@ cdef _validate_4d_list(in_list, list_name):
         )
 
 cpdef np.ndarray _decompress(
-    uint8_t[::1] compressed_data,
+    const uint8_t[::1] compressed_data,
     zfp_type ztype,
     shape,
     out=None,
@@ -261,10 +261,10 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = <char*>&compressed_data[0]
+    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        comp_data_pointer,
+        <char*>comp_data_pointer,
         len(compressed_data)
     )
     cdef zfp_stream* stream = zfp_stream_open(bstream)
@@ -330,15 +330,15 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-    uint8_t[::1] compressed_data,
+     uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = <char *>&compressed_data[0]
+    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        comp_data_pointer,
+        <char *>comp_data_pointer,
         len(compressed_data)
     )
     cdef zfp_stream* stream = zfp_stream_open(bstream)
--- a/tests/python/test_numpy.py
+++ b/tests/python/test_numpy.py
@@ -59,12 +59,16 @@ class TestNumpy(unittest.TestCase):
                 compress_param_num
             ),
         }
-        compressed_array = zfpy.compress_numpy(
+        compressed_array_t = zfpy.compress_numpy(
             random_array,
             write_header=False,
             **compression_kwargs
         )
-
+        if isinstance(compressed_array_t, np.ndarray):
+            compressed_array = compressed_array_t
+        else:
+            mem = memoryview(compressed_array_t)
+            compressed_array = np.array(mem, copy=False)
         # Decompression using the "advanced" interface which enforces no header,
         # and the user must provide all the metadata
         decompressed_array = np.empty_like(random_array)
@@ -177,11 +181,16 @@ class TestNumpy(unittest.TestCase):
                             # Decompression using the "public" interface
                             # requires a header, so re-compress with the header
                             # included in the stream
-                            compressed_array = zfpy.compress_numpy(
+                            compressed_array_t = zfpy.compress_numpy(
                                 random_array,
                                 write_header=True,
                                 **compression_kwargs
                             )
+                            if isinstance(compressed_array_t, np.ndarray):
+                                compressed_array = compressed_array_t
+                            else:
+                                mem = memoryview(compressed_array_t)
+                                compressed_array = np.array(mem, copy=False)
                             decompressed_array = zfpy.decompress_numpy(
                                 compressed_array,
                             )
--- a/tests/python/test_numpy.py
+++ b/tests/python/test_numpy.py
@@ -59,16 +59,13 @@ class TestNumpy(unittest.TestCase):
                 compress_param_num
             ),
         }
-        compressed_array_t = zfpy.compress_numpy(
+        compressed_array_tmp = zfpy.compress_numpy(
             random_array,
             write_header=False,
             **compression_kwargs
         )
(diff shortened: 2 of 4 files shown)


python/zfpy.pyx
```diff
@@ -260,7 +261,7 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef char* comp_data_pointer = <char*>&compressed_data[0]
```
jakirkham, line 264:
```suggestion
    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
```

python/zfpy.pyx
```diff
@@ -329,12 +330,12 @@ cpdef np.ndarray _decompress(
...
 
 cpdef np.ndarray decompress_numpy(
-    bytes compressed_data,
+    const uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef char* comp_data_pointer = <char *>&compressed_data[0]
```
jakirkham, line 338:
```suggestion
    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
```

jakirkham (commented):
Thanks @halehawk! Have a couple minor suggestions below.

jakirkham:
It seems [`stream_open`]( https://github.com/LLNL/zfp/blob/2fbe0084a7e08374e2ee199a7d393d565c0c6d82/include/bitstream.h#L18-L19 ) takes a `void*` that is non-`const`. Is there a reason for that @lindstro? Are there cases where it writes back to the data provided? If so, that would be tricky with `bytes` objects even before this change (as they are immutable in Python). How should we go about handling read-only data then?

lindstro:
> It seems [`stream_open`](https://github.com/LLNL/zfp/blob/2fbe0084a7e08374e2ee199a7d393d565c0c6d82/include/bitstream.h#L18-L19) takes a `void*` that is non-`const`. Is there a reason for that @lindstro? Are there cases where it writes back to the data provided? If so, that would be tricky with `bytes` objects even before this change (as they are immutable in Python). How should we go about handling read-only data then?

@jakirkham The reason is that the bit stream is used both for compression and decompression, and it needs to be mutable to support compression.

jakirkham:
Does that mean user provided data is overwritten as part of compression?

lindstro:
> Does that mean user provided data is overwritten as part of compression?

No, user data is not being overwritten.  The compressed data is copied and returned here: https://github.com/LLNL/zfp/blob/0817b0865e13f54af573bf56040fee5869ea4cbd/python/zfpy.pyx#L166.  I'm unsure how this all relates to `stream_open`, which is a very low-level C function that is called only once during compression to set up the memory buffer where the compressed data is written: https://github.com/LLNL/zfp/blob/0817b0865e13f54af573bf56040fee5869ea4cbd/python/zfpy.pyx#L155.

I'm not entirely sure I understand the issue being discussed here.  Can you please elaborate?

jakirkham:
A `bytes` object is read-only. So the pointer would need to be `const`. It may have worked before with a C compiler warning. Though this change is highlighting that is a problem as Cython is making this a compilation error now. Would it be possible to provide API functions that are `const` at least in some cases to respect this constraint? I'm guessing this means other changes under-the-hood as well.

halehawk:
stream_open is used in both compress and decompress, so its argument has to
be writable.

lindstro:
Supporting both const and nonconst pointers in the bit stream API would require substantial surgery on all of zfp.  Moreover, it would preclude you from compressing to a stream and later decompressing from the same stream (something needed by zfp's compressed-array classes)--two different objects would be needed: one that holds a const pointer, one that holds a mutable pointer.  From the C/C++ programmer's perspective, the expectation is that `stream_open` is initialized with a pointer resulting from `malloc`, `new`, or similar, and that this memory buffer must be populated with compressed data at some point, which requires write access.

Now, we should be able to safely cast a `bytes` or `memoryview` object to `const void*`, no?  We can then use another cast in the call to `stream_open` to do away with the const qualifier.

jakirkham:
Good to know.

Sure we could do that. Just noting that we may want to fix this in the long term.

@halehawk, would you be able to give Peter's suggestion a try? Should be a small change to make.

halehawk:
I just added stream_open(<char *>buffer, ..), is this the way you want?
Then after I submitted the change, it is still failed. @lindstro you can
check the error log of Xenial. @jakirkham

lindstro:
For some reason, build #992 is not showing up on CDash (you may be able to check it yourself: https://open.cdash.org/index.php?project=zfp), even though the Travis logs suggest that they were sent there.  If they don't show up on CDash soon, I may restart the build.

Your changes look right, although I would caution against using `char*` in place of `unsigned char*` or, preferably, `void*`.

Although mostly orthogonal to this issue, the compressed stream needs to be [word aligned](https://zfp.readthedocs.io/en/release0.5.5/installation.html#c.BIT_STREAM_WORD_TYPE). I'm not sure if Python makes any guarantees about alignment.  I have no reason to believe that alignment has anything to do with these errors; this is just something that occurred to me when discussing pointer types.

lindstro:
Looks like some tests but not all are now passing:

```
test_advanced_decompression_checksum (__main__.TestNumpy) ... ok
test_advanced_decompression_nonsquare (__main__.TestNumpy) ... ok
test_different_dimensions (__main__.TestNumpy) ... ERROR
test_different_dtypes (__main__.TestNumpy) ... ERROR
test_utils (__main__.TestNumpy) ... ERROR

======================================================================
ERROR: test_different_dimensions (__main__.TestNumpy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_numpy.py", line 24, in test_different_dimensions
    self.lossless_round_trip(c_array)
  File "test_numpy.py", line 17, in lossless_round_trip
    decompressed_array = zfpy.decompress_numpy(compressed_array)
  File "zfpy.pyx", line 333, in zfpy.decompress_numpy
  File "stringsource", line 658, in View.MemoryView.memoryview_cwrapper
  File "stringsource", line 349, in View.MemoryView.memoryview.__cinit__
BufferError: Object is not writable.

======================================================================
ERROR: test_different_dtypes (__main__.TestNumpy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_numpy.py", line 38, in test_different_dtypes
    self.lossless_round_trip(array)
  File "test_numpy.py", line 17, in lossless_round_trip
    decompressed_array = zfpy.decompress_numpy(compressed_array)
  File "zfpy.pyx", line 333, in zfpy.decompress_numpy
  File "stringsource", line 658, in View.MemoryView.memoryview_cwrapper
  File "stringsource", line 349, in View.MemoryView.memoryview.__cinit__
BufferError: Object is not writable.

======================================================================
ERROR: test_utils (__main__.TestNumpy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_numpy.py", line 186, in test_utils
    compressed_array,
  File "zfpy.pyx", line 333, in zfpy.decompress_numpy
  File "stringsource", line 658, in View.MemoryView.memoryview_cwrapper
  File "stringsource", line 349, in View.MemoryView.memoryview.__cinit__
BufferError: Object is not writable.

----------------------------------------------------------------------
Ran 5 tests in 17.482s

FAILED (errors=3)
```

python/zfpy.pyx
```diff
@@ -329,15 +330,15 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-    bytes compressed_data,
+     uint8_t[::1] compressed_data,
```
jakirkham, line 333:
```suggestion
    const uint8_t[::1] compressed_data,
```

jakirkham: Should fix the test error

python/zfpy.pyx
```diff
@@ -260,10 +261,10 @@ cpdef np.ndarray _decompress(
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        comp_data_pointer,
+        <char*>comp_data_pointer,
```
jakirkham, line 267:
```suggestion
        <void*>comp_data_pointer,
```

python/zfpy.pyx
```diff
@@ -329,15 +330,15 @@ cpdef np.ndarray _decompress(
...
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        comp_data_pointer,
+        <char *>comp_data_pointer,
```
jakirkham, line 341:
```suggestion
        <void *>comp_data_pointer,
```

halehawk:
Xenial of 992 builds still gave us 5 errors. The less errors log Peter
showed were built from non-xenial builds.
My 993 builds with all suggested changes still failed on xenial.

lindstro:
Sorry, I copied the wrong log.  I'm guessing this is a Cython bug (see, e.g., https://github.com/cython/cython/issues/1605). I don't know what to suggest at this point.

@jakirkham Do you have other numcodecs examples that use Cython with `memoryviews` successfully?

jakirkham:
Though that has been fixed since 0.28. Cython is now deep in the 0.29.x releases.

python/zfpy.pyx
```diff
@@ -329,15 +330,15 @@ cpdef np.ndarray _decompress(
     return output
 
 cpdef np.ndarray decompress_numpy(
-    bytes compressed_data,
+     const uint8_t[::1] compressed_data,
```
jakirkham, line 333: Sorry there may have been an extra space in the suggestion before. Maybe that helps?

```suggestion
    const uint8_t[::1] compressed_data,
```

jakirkham:
Where do we look to see the build failures? Went to the CDash page, but am not familiar with this tool. So wasn't sure where to go from there to see this PR.

lindstro:
The CDash user interface leaves much to be desired.  You first need to find the correct date of the build (in whatever timezone CDash defaults to), which you can do via current/prev/next (e.g., https://open.cdash.org/index.php?project=zfp&date=2020-09-18).  Then find the Travis build name you're interested in (e.g., develop-#993.1).  Click the entry in the red Fail column (e.g., https://open.cdash.org/viewTest.php?onlyfailed&buildid=6770400), then the "Failed" link (e.g., https://open.cdash.org/test/250854826).  This should bring up the error log for that build.

jakirkham:
Thanks. Can see the error now.

jakirkham:
Would it be possible to add `cython --version` in the Travis build somewhere? Just to confirm we are not picking up an old version.

lindstro:
Somewhere in travis.sh should do the job.

halehawk:
In xenial, cython version is 0.23.

halehawk:
See cython in the attached screenshot.

halehawk:
That's because a lot of python packages are not developed with Python3.5
any more, such as I only can get numcodecs 0.5.5 in my python3.5
environment.

halehawk:
Can we enforce a newer version of cython?

lindstro:
Would `sudo $PYTHON_EXECUTABLE -m pip install --upgrade cython` work (see [.travis.yml](https://github.com/LLNL/zfp/blob/0817b0865e13f54af573bf56040fee5869ea4cbd/.travis.yml#L284))?

jakirkham:
We could also create a Conda or virtualenv based environment with newer dependencies.

jakirkham:
Also we are discussing dropping Python 3.5 in Numcodecs and Zarr (in case anyone has thoughts on that).

xref: https://github.com/zarr-developers/governance/issues/11

halehawk:
Good news, after I successfully installed cython0.29, it passed some
python3.5 xenial tests 1003.3 and 1003.13. I assume it shall pass all
python3.5 xenial tests now. I need to add cython update in python2 now.

halehawk:
There are no python2 builds on xenial machine. And three failed builds are
failed on uploading the reports not the test_numpy results.
So we got this passed. Thank you very much.
@lindstro. do I need to run again to get every build passed?

lindstro:
@halehawk I wouldn't worry about python2--we should probably drop python2 support since the language is officially obsolete.

The failed CDash uploads are maddening--more often than not, they cause otherwise successful tests to fail.  I restarted the Travis jobs to make sure everything passes.

python/zfpy.pyx
```diff
@@ -245,25 +246,25 @@ cdef _validate_4d_list(in_list, list_name):
...
+    const uint8_t[::1] compressed_data,
     zfp_type ztype,
     shape,
     out=None,
     double tolerance = -1,
     double rate = -1,
     int precision = -1,
 ):
-
+    print(cython.__version__)
```
jakirkham, line 257: I think we can drop this now

jakirkham:
Nicely done! 😄

halehawk:
Yes. I will remove it.

lindstro:
@halehawk Looks like Travis tests are now passing--well done!  For the benefit of our users, it would be great to know what minimal Cython version is needed so we can include that with the documentation.  Were you able to determine when the Cython bug was fixed?

halehawk:
I am supposed that cython should be 0.28.5 or later (I tested this version
on my machine). But in our cases, the cython 0.29.21 is installed by pip.

halehawk:
Why do appveyor tests take so long? I hope this PR can be merged soon, and
you can release a new pip package. Thus I can use from numcodes.zfpy, and
make it merge too.

lindstro:
AppVeyor runs for every commit you make, but unlike Travis it runs each job sequentially, and each job takes up to 30 minutes.  10 builds x 12 jobs x 30 minutes = 60 hours.

I canceled all builds other than your last commit.  This test should finish later today.

I'm a little hesitant to releasing a pip package before the next zfp release.  We've been synchronizing pip package versions with zfp releases, and our plan is to release zfp 0.5.6 in the next few weeks.  Can this wait until then?

Before we release another pip package, I also want to figure out how to make `delocate` include the zfp dylib with the pip package on macOS.  You had some thoughts on this: https://github.com/zarr-developers/numcodecs/pull/229#issuecomment-639947775.  But it wasn't clear to me what changes need to be made.  Can you please follow up by filing a new zfpy-wheels issue so this can be discussed there?

halehawk:
Is it possible you release a minor release such as 0.5.5.1 (-:?
If not, please merge my PR as soon as possible.
I will be glad to help on this dylib issue. But can you point to me how you
can run the release script?

halehawk:
Do I need to do anything with code coverage?

jakirkham:
It might be good to have a test using a NumPy array to confirm that works. Though I don't think that is related to the line coverage report per se.

halehawk:
I did a test on my machine with ensure_ndarray, it worked. But do I need to
do it officially on test_numpy, ithen test_numpy.py needs import numcodecs.

jakirkham:
Sure I think it would be useful to have it included in zfp to ensure the intended behavior continues to work going forward.

python/zfpy.pyx
```diff
@@ -245,25 +246,24 @@ cdef _validate_4d_list(in_list, list_name):
...
 ):
-
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
     if compressed_data is out:
         raise ValueError("Cannot decompress in-place")
     _validate_4d_list(shape, "shape")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef const char* comp_data_pointer = <const char*>&compressed_data[0]
```
lindstro, line 263: Can we please change this to `const void*` to avoid any confusion why the pointer is initially cast to a (signed) `char*`?

python/zfpy.pyx
```diff
@@ -245,25 +246,24 @@ cdef _validate_4d_list(in_list, list_name):
...
     cdef zfp_field* field = zfp_field_alloc()
     cdef bitstream* bstream = stream_open(
-        comp_data_pointer,
+        <void *>comp_data_pointer,
```
lindstro, line 266: Is this second cast from `const void*` to `void*` necessary, or can `comp_data_pointer` be declared as `void*`?

python/zfpy.pyx
```diff
@@ -329,15 +329,15 @@ cpdef np.ndarray _decompress(
...
 
 cpdef np.ndarray decompress_numpy(
-    bytes compressed_data,
+     const uint8_t[::1] compressed_data,
 ):
     if compressed_data is None:
         raise TypeError("compressed_data cannot be None")
 
-    cdef char* comp_data_pointer = compressed_data
+    cdef const char* comp_data_pointer = <const char *>&compressed_data[0]
```
lindstro, line 337: Same thing here; `const void*` or `void*`.

lindstro:
> Is it possible you release a minor release such as 0.5.5.1 (-:? If not, please merge my PR as soon as possible. I will be glad to help on this dylib issue. But can you point to me how you can run the release script?
> […](#)

@halehawk Thanks for the offer.  By the release script, do you mean generating and uploading the wheels to PyPI?  That's done through a separate repo: https://github.com/LLNL/zfpy-wheels.  When experimenting with this, we should use TestPyPI; see https://packaging.python.org/guides/using-testpypi/.  If this is what you meant, then let's discuss offline the steps involved in doing this.

lindstro:
> I did a test on my machine with ensure_ndarray, it worked. But do I need to do it officially on test_numpy, ithen test_numpy.py needs import numcodecs.
> […](#)

I'm not sure I follow.  Why is numcodecs needed to test that ndarrays can be passed?  I'm reluctant to add unnecessary dependencies to zfp.

halehawk:
Sure, I will change as you mentioned. Because your codes used char* at
first, I thought I should follow it.

halehawk:
Yes, that's what I want to know.

halehawk:
Yes, that's my concern too. I will try to test directly with a numpy array.

halehawk:
I added tests and passed all checks. @lindstro

tests/python/test_numpy.py
```diff
@@ -59,12 +59,16 @@ def test_advanced_decompression_checksum(self):
...
             ),
         }
-        compressed_array = zfpy.compress_numpy(
+        compressed_array_t = zfpy.compress_numpy(
             random_array,
             write_header=False,
             **compression_kwargs
         )
-
+        if isinstance(compressed_array_t, np.ndarray):
```
lindstro, line 67: I'm not sure I understand the intent here.  `compressed_array_t` is by definition a `bytes` object, so this conditional should always be false.

Wouldn't the appropriate test be to ensure that we can pass both a `bytes` object and an `ndarray` to `zfpy.decompress_numpy`?  What other array-like objects are we trying to support here?

Minor nit: As a C/C++ programmer, I think `compressed_array_t` is a confusing name as the `_t` suffix makes it look like a type.

halehawk:
On Wed, Sep 30, 2020 at 12:09 PM Peter Lindstrom <notifications@github.com>
wrote:

> *@lindstro* commented on this pull request.
> ------------------------------
>
> In tests/python/test_numpy.py
> <https://github.com/LLNL/zfp/pull/106#discussion_r497705642>:
>
> >              random_array,
>              write_header=False,
>              **compression_kwargs
>          )
> -
> +        if isinstance(compressed_array_t, np.ndarray):
>
> I'm not sure I understand the intent here. compressed_array_t is by
> definition a bytes object, so this conditional should always be false.
>
> No, compressed_array_t could be array objects from numcodecs.zfpy, and
could be a bytes object from your other  upstream application tools.

> Wouldn't the appropriate test be to ensure that we can pass both a bytes
> object and an ndarray to zfpy.decompress_numpy? What other array-like
> objects are we trying to support here?
>
Yes, in your test_numpy.py, there are two places calling _decompress (line
75 and line 104) or decompress_numpy, I only added this test on one place
of _decompress or decompress_numpy.

lindstro:
> No, compressed_array_t could be array objects from numcodecs.zfpy, and could be a bytes object from your other upstream application tools.

Maybe it's my lack of Python skills, but unless I'm missing something obvious, in this test `compressed_array_t` is always a `bytes` object returned by `zfpy.compress_numpy`.  How could it be any other type the way this particular test is written?

>> Wouldn't the appropriate test be to ensure that we can pass both a bytes object and an ndarray to zfpy.decompress_numpy? What other array-like objects are we trying to support here?

> Yes, in your test_numpy.py, there are two places calling _decompress (line 75 and line 104) or decompress_numpy, I only added this test on one place of _decompress or decompress_numpy.

My concern is that the conditional is always false (so what purpose does it serve?), and therefore only one branch is taken and tested.  That is, only the `memoryview` path is exercised.

halehawk:
I will remove one condition.

jakirkham:
Thanks Haiying! Also thanks Peter for the review! 😄

lindstro:
Yes, big thanks to @halehawk for this PR and for putting up with my picky comments. :wink:   Some of the tests are failing on Travis and AppVeyor, but this appears to be caused by some issues with cmocka.  I've restarted the failed jobs, and they seem to be passing now.

halehawk:
Thank you all for such good collaboration. @lindstro @jakirkham

jakirkham:
Thanks again for your help, Peter! Do you have an idea of when the next release (with this change) might be? 🙂

lindstro:
We're woefully behind schedule on this release, and I don't want to make any promises, but we're working hard to get it done by the end of the year.

halehawk:
Peter, will the compress and decompress APIs in your new release have changes?

lindstro:
No, no changes there.  The majority of new features are related to the C++ compressed-array classes and their C wrappers.

halehawk:
Peter, any update on your new zfp release?

lindstro:
We're slowly making progress, but we're not as close as I would have wished.  I expect we're still a few weeks away.

jakirkham:
Just checking in here, how are things looking?

halehawk:
Also, it looks like numcodecs latest release including Python 3.9
builds. @lindstro could you please build zfpy with this Python version as
well?

lindstro:
> Just checking in here, how are things looking?

Slowly making progress on 0.5.6.  Sorry to be holding everyone up, but there's just a lot of stuff going into this release.

Regarding Python 3.9, I can add that to zfpy-wheels but am unsure whether it needs `MB_ML_VER=2010`, like 3.8.

@da-wad Do you know?

jakirkham:
No worries 🙂

If it is easy to add, it would be great, but if not wouldn't worry about it. I don't think the fact that Numcodecs uses Python 3.9 should constrain zfpy.

Ah guessing you are referring to [this]( https://github.com/matthew-brett/multibuild#build-phase )? Yeah 2010 makes sense sense as that is CentOS 6 based. I think the default is 1, which is CentOS 5 based (no one should be using that any more as it no longer receives any updates). They've also added 2014, which is CentOS 7 based. Not sure what GLIBC your users typically need, but continuing to target CentOS 6 makes sense and should give GLIBC 2.12+ support.

jakirkham:
We need a tag is the issue

halehawk:
On Fri, Mar 12, 2021 at 12:09 PM jakirkham ***@***.***> wrote:

> We need a tag is the issue
>
I didn't get it, could you please elaborate it?

jakirkham:
There is no git tag here including these changes from which to build. We would need that to proceed (even with Conda)

PR #109, opened 2020-10-12
Avoid importing unused cpython array for compatibility with pypy

hmaarrfk:
Do you have a changelog you want me to add a note in?

A good explanation as to why this is important
https://github.com/h5py/h5py/issues/1514

### DIFF
--- a/python/zfpy.pyx
+++ b/python/zfpy.pyx
@@ -4,9 +4,7 @@ import functools
 import cython
 from libc.stdlib cimport malloc, free
 from cython cimport view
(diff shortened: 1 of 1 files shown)


lindstro:
We keep the changelog locally.  If you write a short note, I will include it with the next release.

Thanks for identifying this issue.  I will merge once the tests pass.

hmaarrfk:
I think the title says it all. Thanks!

GarrettDMorrison:
LGTM

PR #113, opened 2020-12-01
Feature/parallel variable rate decompression

LennartNoordsij:
Initial commit of parallel variable rate decompression

- Added zfp_index struct
- Added several types and functions needed for the zfp index
- Added openmp decompression for variable and fixed rate
- Added CUDA decompression for variable rate
- Cleaned up the CUDA code
- Other changes

This version supports encoding of the following index types:
- Offset (with granularity)
- Hybrid (with granularity)
- Length

This version is a work in progress, it has been verified with some test datasets but does not yet contain unit tests.


lindstro:
@LennartNoordsij Thanks for this significant contribution!  Some of the (existing?) OpenMP tests are failing on Travis.  Can you please investigate?

LennartNoordsij:
Hi Peter,

The test that I think is failing is the following:
"given_withoutOpenMP_whenDecompressOmpPolicy_expect_noop" 
It expects no operation when attempting decompression with OpenMP. However this patch implements OpenMP decompression, so it actually decompresses the stream instead of no operation. It is expected that this check will fail, here the unit tests should be extended to cover the new functionality.

lindstro:
That would be surprising since this test is run only when OpenMP is [fully disabled](https://github.com/LLNL/zfp/blob/1829523e6ffbc9d42fa4467d7bee9d30735d9608/tests/src/execPolicy/testOmp.c#L209) at compile time.  You also don't want the test to pass when zfp has been configured without OpenMP support.

Can you run `ctest -R Omp -V` to pinpoint which test(s) fail?

LennartNoordsij:
Hi Peter,

Actually, it is also run [when OpenMP is enabled](https://github.com/LLNL/zfp/blob/1829523e6ffbc9d42fa4467d7bee9d30735d9608/tests/src/execPolicy/testOmp.c#L203) at compile time. In that scenario, this test fails. In the cases where OpenMP is disabled at compile time you can see that all of the tests pass in the travis reports.

Running the test I get the following result:
[==========] Running 6 test(s).
[ RUN      ] given_withOpenMP_when_setExecutionOmp_expect_set
[       OK ] given_withOpenMP_when_setExecutionOmp_expect_set
[ RUN      ] given_withOpenMP_when_setOmpThreads_expect_set
[       OK ] given_withOpenMP_when_setOmpThreads_expect_set
[ RUN      ] given_withOpenMP_serialExec_when_setOmpThreads_expect_setToExecOmp
[       OK ] given_withOpenMP_serialExec_when_setOmpThreads_expect_setToExecOmp
[ RUN      ] given_withOpenMP_when_setOmpChunkSize_expect_set
[       OK ] given_withOpenMP_when_setOmpChunkSize_expect_set
[ RUN      ] given_withOpenMP_serialExec_when_setOmpChunkSize_expect_setToExecOmp
[       OK ] given_withOpenMP_serialExec_when_setOmpChunkSize_expect_setToExecOmp
[ RUN      ] given_withOpenMP_whenDecompressOmpPolicy_expect_noop
[  ERROR   ] --- EXCEPTION_ACCESS_VIOLATION occurred at 79D00316.
...
[  FAILED  ] given_withOpenMP_whenDecompressOmpPolicy_expect_noop
[==========] 6 test(s) run.
[  PASSED  ] 5 test(s).
[  FAILED  ] 1 test(s), listed below:
[  FAILED  ] given_withOpenMP_whenDecompressOmpPolicy_expect_noop

Which I think confirms my earlier statement. Does this cover your concerns?

lindstro:
Oh, but that's a different test (`given_withOpenMP_whenDecompressOmpPolicy_expect_noop != given_withoutOpenMP_whenDecompressOmpPolicy_expect_noop`).  Since OpenMP decompression is now supported, we should remove the first test and add some new ones.  We can write the new tests, but I would like the PR to be clean and pass before merging, so can you please remove `given_withOpenMP_whenDecompressOmpPolicy_expect_noop` and resubmit?

GarrettDMorrison:
LGTM

PR #130, opened 2021-05-12
CUDA variable bitrate

gthomascollignon:
Work in progress.
Compression with variable bit rate seems to work on GPU, but needs more testing.
The individual lengths of each block are not saved unless the user provides a pointer through the bitstream struct (need a way to set it).

The variable bit rate requires CUDA 9 (cooperative launch and grid sync, to compact the data in-place). Need to hide the variable bit rate code for CUDA <9, or bump up the min requirement to 9?
The code uses CUB which is included in CUDA 11+, but available otherwise (todo: update cmake to download it).
If CUDA 11.2 is found, the temporary memories on GPU are allocated with cudaMallocAsync.


gthomascollignon:
Closing. Moved to feature/cuda_variable_rate.

PR #132, opened 2021-05-13
Feature/cuda variable rate

gthomascollignon:
Zero-padded the output stream.
Added explicit architectures in cmake
Disabled unsupported reversible mode on GPU.
Bumped up min CUDA version to 9.


lindstro:
LGTM

PR #137, opened 2021-06-09
CMake: (fix) OpenMP dependency.

jwsblokland:
Hello,

In the original implementation the OpenMP libraries has been added
to the INTERFACE_LINK_LIBRARIES in the zfp-targets.cmake file using
an absolute path. This works but it is not portable. Using CMake
version 3.9 and latter the find_package(OpenMP ...) has been
significantly improved which allows for a relocatable solution of the
CMake config files. In addition, added ZFP_WITH_OPENMP variable to the
config file zfp-config.cmake to indicate if the zfp library has
been built with OpenMP support. If true, load the OpenMP package
using find_package(OpenMP ...).

Maybe a similar thing can be done for CUDA. Unfortunately, I do not 
have access to both CUDA and nVidia GPU therefore I cannot investigate
how this should be done in your CMake configuration.

Best regards,
   Jan-Willem

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,4 @@
-if(WIN32)
-  cmake_minimum_required(VERSION 3.4)
-else()
-  cmake_minimum_required(VERSION 3.1)
-endif()
+cmake_minimum_required(VERSION 3.9)
 
 # Fail immediately if not using an out-of-source build
 if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,11 +1,11 @@
 if(ZFP_WITH_CUDA)
-  SET(CMAKE_CXX_FLAGS_PREVIOUS ${CMAKE_CXX_FLAGS})
-  SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fPIC" )
+  set(CMAKE_CXX_FLAGS_PREVIOUS ${CMAKE_CXX_FLAGS})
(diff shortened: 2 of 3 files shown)


jwsblokland:
I noticed osx 2 tests of the Travis CI build failed. My guess that the reason they fail is due to fact Xcode 7.3 has been used. According to the Travis documentation this Xcode version makes use of CMake 3.5.2 which lower than the required CMake version 3.9.

Now the question is what is more important for you. Having a relocatable solution of the CMake config files or keep osx tests with Xcode 7.3 working.

lindstro:
You're right:

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.9 or higher is required.  You are running version 3.6.2
```

It's probably OK to require CMake 3.9, but we'll have to update our Travis build matrix, which is something we're already in the process of doing.  Let's put this PR on hold for now until we're done with updates on our end.

jwsblokland:
I am glad to hear this CMake 3.9 requirement will most likely not be a problem. 

Is it possible to give me an estimation when your updates are done? This gives me the possibility to manage the expectations of my colleagues.

lindstro:
This may be a bigger issue than I first thought as we're debating whether to move off Travis CI altogether.  That said, I think we can drop the Xcode 7 tests now just to get your PR to pass cleanly, which shouldn't take more than a day or so to address.  Of course, this fix will limited to the `develop` branch until the next release, which likely is several weeks away.

jwsblokland:
The `develop` branch is fine me. I can easy explain it to them that you have your own release cycle just like the software we here at Shell are deploying.

lindstro:
I dropped xcode7.3 from our Travis build matrix on `develop`.  Can you please rebase your PR?

jwsblokland:
Thanks. Sure no problem to rebase to the latest status of the `develop` branch.

lindstro:
LGTM

PR #139, opened 2021-07-19
Feature/cuda variable rate

gthomascollignon:
- Fixed bug that could create invalid checksums (from zero-padded area at the very end). The size wasn't rounded properly to the next multiple of 64b.
- Added missing validity check for first stream of the block, this could create out of bounds memory accesses.

### DIFF
--- a/src/cuda_zfp/cuZFP.cu
+++ b/src/cuda_zfp/cuZFP.cu
@@ -556,7 +556,13 @@ cuda_compress(zfp_stream *stream, const zfp_field *field, int variable_rate)
     size_t blocks = zfp_field_num_blocks(field);
     for (size_t i = 0; i < blocks; i += chunk_size)
     {
-      int cur_blocks = i + chunk_size <= blocks ? chunk_size : (int)(blocks - i);
+      int cur_blocks = chunk_size;
+      bool last_chunk = false;
+      if (i + chunk_size > blocks)
+      {
+        cur_blocks = (int)(blocks - i);
+        last_chunk = true;
+      }
       // Copy the 16-bit lengths in the offset array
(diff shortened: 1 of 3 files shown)


src/cuda_zfp/cuZFP.cu
```diff
@@ -580,8 +580,9 @@ cuda_compress(zfp_stream *stream, const zfp_field *field, int variable_rate)
       cuZFP::chunk_process_launch((uint*)d_stream, d_offsets, i, cur_blocks, last_chunk, buffer_maxbits, num_sm);
     }
     // The total length in bits is now in the base of the prefix sum.
+    // but it excludes the 64-bit "add_padding". -> Pad to 64b, convert to bytes
     cudaMemcpy (&stream_bytes, d_offsets, sizeof (unsigned long long), cudaMemcpyDeviceToHost);
-    stream_bytes = (stream_bytes + 7) / 8;
+    stream_bytes = ((stream_bytes + 63) / 64) * 8;
```
lindstro, line 585: I don't think this was wrong in the first place.  This rounds the number of *bytes* (not bits) to the next multiple of 8 bytes = 64 bits.

lindstro: Sorry, I should have looked at the code above for context.  I guess `stream_bytes` initially is measured in number of bits, so this does do the correct thing.

gthomascollignon: Correct, the comment states the offset is still in bits. I could have use 2 variables stream_bits and stream_bytes to make it more explicit.

lindstro: That being said, `stream_bytes` is not a great name for a bit offset.  A more appropriate name may promote code readability.

lindstro:
LGTM

PR #145, opened 2021-10-04
Add memset for bit stream

JieyangChen7:
Add memset function for bit stream for fixing a bug in the HIP version.

### DIFF
--- a/utils/zfp.c
+++ b/utils/zfp.c
@@ -534,7 +534,10 @@ int main(int argc, char* argv[])
     }
 
     /* compress data */
-    zfpsize = zfp_compress(zfp, field);
+    for (int i = 0; i < 5; i++) {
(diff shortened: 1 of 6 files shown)


utils/zfp.c
```diff
@@ -534,7 +534,10 @@ int main(int argc, char* argv[])
     }
 
     /* compress data */
-    zfpsize = zfp_compress(zfp, field);
+    for (int i = 0; i < 5; i++) {
```
lindstro, line 537: This is useful for profiling but should not be the default behavior in the zfp command-line utility.

JieyangChen7: Thanks for the reminder. That commit was just for profiling. I'm closing this PR for now to avoid confusion.

PR #148, opened 2021-10-29
CMAKE: ZFP_INCLUDE_DIRS avail at configure time

vicentebolea:
Similarly to #147 this PR makes ZFP_INCLUDE_DIRS available at configure time (before at generator time).

This is needed for projects that depends on zfp.


lindstro:
@vicentebolea Thanks for another PR.  I'm afraid I don't know CMake well enough to tell how this change might impact our users.  Before we merge this PR, I'd like to know better what issue it solves.

I'd also welcome feedback from @chuckatkins, who wrote the original config.

vicentebolea:
@lindstro CMake has two main runtime phases, the first is the easiest one and is the one that happens at first during "Configuring". In this case CMake does works as a normal imperative interpreted language.

Once this phase is done and all its targets are "configured".  Cmake will write the Makefile, build.ninja or any other generator files. This is the generator phases, at this time are those generator-expression computed (The ones with the form `$<$<>:>`) and written as they are in those generator files. This is great since those expressions gives us lots of flexibility.

The problem is is that when you need the value of variable defined in the form of  `$<$<>:>` at _configure_ time.

lindstro:
LGTM

PR #156, opened 2022-02-28
Update setup.py to v1.0.0

da-wad:
The new version number should be in place before making a v1.0.0 release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import numpy as np
 
 setup(
     name="zfpy",
-    version="0.5.5",
+    version="1.0.0",
     author="Peter Lindstrom",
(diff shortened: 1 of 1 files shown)


lindstro:
@da-wad FYI, the release of zfp 1.0.0 is imminent.  I forgot that you had submitted a PR--setup.py has already been updated on the release1.0.0 branch, where that kind of change ought to be made anyway.  I will close this PR (without merging), but please let us know if you need anything else from us for the release.  Would be great if you could take care of PIP packages once the release is official.

da-wad:
Not a problem, I see that @GarrettDMorrison made exactly this change. I've been working on the pip packages today and will update you!

PR #161, opened 2022-04-11
Fix build system for ROCM 5.1.0

JieyangChen7:


CMakeLists.txt
```diff
@@ -190,36 +190,8 @@ if(ZFP_WITH_CUDA)
...
-    set(HIP_PATH $ENV{HIP_PATH} CACHE PATH "Path to which HIP has been installed")
-  endif()
-  set(CMAKE_MODULE_PATH "${HIP_PATH}/cmake" ${CMAKE_MODULE_PATH})
-  find_package(HIP REQUIRED)
-  if(HIP_FOUND)
-    message(STATUS "Found HIP(${HIP_VERSION}) in ${HIP_PATH}")
-  else()
-    message(FATAL_ERROR "Could not find HIP. Ensure that HIP is either installed in /opt/rocm/hip or the variable HIP_PATH is set to point to the right location.")
-  endif()
+  enable_language(HIP)
```
lindstro, line 193: This is clearly a lot cleaner, but enable_language(HIP) is available only as of CMake 3.21.  zfp currently requires CMake 3.9.

We have gone to great lengths to make zfp easy to build, with as few requirements and dependencies as possible, and with conformance to old C89 and C++98.  We need to consider the implications of jumping from CMake 3.9 to 3.21 and from C++98 to C++14 just to support HIP.

JieyangChen7: @lindstro Due to the changes in ROCM 5.1.0, we can no longer build ZFP with the existing CMake build system and C++98. This is because ROCM 5.1.0 starts to rely on features in C++14 (e.g., using `std::conditional_t` type). Also, when using the old way (currently used in ZFP) to handle HIP code in CMake (i.e., using `set_source_files_properties(${zfp_source} PROPERTIES HIP_SOURCE_PROPERTY_FORMAT 1)`), the `-std=c++14` flag cannot be correctly passed to the `hipcc` compiler, since CMake cannot correctly recognize the source code language. To let CMake know our source code is written in HIP, we have to `enable_language(HIP)` and `set_source_files_properties(${HIPZFP_SOURCE} PROPERTIES LANGUAGE HIP)`. So, it seems that using `enable_language(HIP)` is our only option to let ZFP build with the latest ROCM. Perhaps it is better to conditional require a newer CMake and C++ versions? For example, we can conditional require CMake 3.21+ and C++14 (or newer) only if users want to build ZFP with HIP.

lindstro: Is C++14 needed even if you don't use such features, e.g., do the headers not compile with pre C++14 compilers?

I agree that we should conditionally require CMake 3.21 and C++14 only when HIP support is enabled.  Can you update the PR to do so?

We should also be better about testing builds with the oldest version of CMake required to catch the use of unsupported CMake features.  That's on us.

JieyangChen7: Those C++14 features are not used in the implementation of ZFP. They are actually used in the HIP runtime system (starting from v5.1.0). So, C++14 is required if we include the HIP runtime header in ZFP. We have no control on whether or not to use those C++14 features. If we compile with pre C++14 compilers, the compiler will complain that the definition of `std::conditional_t` type cannot be found. I will update the PR to add conditional requirements in CMake.

lindstro:
While I can get zfp built on Crusher using ROCM 5.1, this PR breaks support for 5.0 and earlier:
```
cmake .. -DZFP_WITH_HIP=ON -DZFP_WITH_OPENMP=OFF -DCMAKE_CXX_FLAGS="--amdgpu-target=gfx90a" -DCMAKE_C_COMPILER=hipcc -DCMAKE_CXX_COMPILER=hipcc
-- The C compiler identification is Clang 14.0.0
-- The CXX compiler identification is Clang 14.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/rocm-5.0.2/hip/bin/hipcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/rocm-5.0.2/hip/bin/hipcc - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Compiling with C standard: 99
-- Compiling with C++ standard: 14
-- The HIP compiler identification is Clang 14.0.0
-- Detecting HIP compiler ABI info
-- Detecting HIP compiler ABI info - failed
-- Check for working HIP compiler: /opt/rocm-5.0.2/llvm/bin/clang++
-- Check for working HIP compiler: /opt/rocm-5.0.2/llvm/bin/clang++ - broken
CMake Error at /autofs/nccs-svm1_sw/crusher/spack-envs/base/opt/linux-sles15-x86_64/gcc-7.5.0/cmake-3.22.1-wd2ptgormfxhkw3fiac75fiznbzcvpah/share/cmake-3.22/Modules/CMakeTestHIPCompiler.cmake:65 (message):
  The HIP compiler

    "/opt/rocm-5.0.2/llvm/bin/clang++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /ccs/home/lindstro/tmp/zfp/build/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/usr/bin/gmake -f Makefile cmTC_a9850/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_a9850.dir/build.make CMakeFiles/cmTC_a9850.dir/build
    gmake[1]: Entering directory '/autofs/nccs-svm1_home1/lindstro/tmp/zfp/build/CMakeFiles/CMakeTmp'
    Building HIP object CMakeFiles/cmTC_a9850.dir/testHIPCompiler.hip.o
    /opt/rocm-5.0.2/llvm/bin/clang++ -D__HIP_ROCclr__=1 -I/opt/rocm-5.0.2/hip/include -I/opt/rocm-5.0.2/include -isystem /opt/rocm-5.0.2/hip/../include -isystem /opt/rocm-5.0.2/llvm/lib/clang/14.0.0 --cuda-host-only  --offload-arch=gfx90a --offload-arch=gfx90a -fPIE -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -o CMakeFiles/cmTC_a9850.dir/testHIPCompiler.hip.o -x hip -c /ccs/home/lindstro/tmp/zfp/build/CMakeFiles/CMakeTmp/testHIPCompiler.hip
    Linking HIP executable cmTC_a9850
    /autofs/nccs-svm1_sw/crusher/spack-envs/base/opt/linux-sles15-x86_64/gcc-7.5.0/cmake-3.22.1-wd2ptgormfxhkw3fiac75fiznbzcvpah/bin/cmake -E cmake_link_script CMakeFiles/cmTC_a9850.dir/link.txt --verbose=1
    /opt/rocm-5.0.2/llvm/bin/clang++  --cuda-host-only  --offload-arch=gfx90a --offload-arch=gfx90a --hip-link CMakeFiles/cmTC_a9850.dir/testHIPCompiler.hip.o -o cmTC_a9850  /opt/rocm-5.0.2/hip/lib/libamdhip64.so.5.0.50002 -lCLANGRT_BUILTINS-NOTFOUND 
    ld.lld: error: unable to find library -lCLANGRT_BUILTINS-NOTFOUND
    clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
    gmake[1]: *** [CMakeFiles/cmTC_a9850.dir/build.make:100: cmTC_a9850] Error 1
    gmake[1]: Leaving directory '/autofs/nccs-svm1_home1/lindstro/tmp/zfp/build/CMakeFiles/CMakeTmp'
    gmake: *** [Makefile:127: cmTC_a9850/fast] Error 2
    
    

  

  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:199 (enable_language)
```
Is this intentional?

Also, all HIP tests fail using ROCM 5.1.  Here's the output of `testexec`:
```
Queue at 0x7fbcabf2a000 inactivated due to async error:
        HSA_STATUS_ERROR_MEMORY_FAULT: Agent attempted to access an inaccessible address
```
I believe this is a different error from the one we've been discussing.  Are you seeing the same?  I may have asked this previously, but is no synchronization needed at some point after the asynchronous memory copy, e.g., at the end of `zfp_compress` and `zfp_decompress`?

CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ if(WIN32)
 else()
   cmake_minimum_required(VERSION 3.1)
```
lindstro, line 4: On the current `develop` branch, we're now requiring CMake 3.9.  Would be good to update it here as well so we do not accidentally revert when we merge later.

JieyangChen7:
> Queue at 0x7fbcabf2a000 inactivated due to async error:
        HSA_STATUS_ERROR_MEMORY_FAULT: Agent attempted to access an inaccessible address

This is the error we have been seeing on MI-250 but not on MI-100. There is no need to add explicit synchronization since we are calling synchronized memory copy API (internally it is implemented with fast asynchronized memory copy and ROCM runtime will add a synchronization for us)
Here is an example of how we call synchronized memory copy API in ZFP:
https://github.com/JieyangChen7/zfp/blob/feature/hip-support/src/hip_zfp/hipZFP.cpp#L347
I have tried to add explicit synchronization but there is no effect.

The `HSA_STATUS_ERROR_MEMORY_FAULT` error is caused by the `concat_bitstreams_chunk` kernel when doing variable rate compression. I suspect it may have something to do with the Cooperative Group since I have encountered bugs in HIP related to Cooperative Group before when I was working on MGARD and HIP. In the end, I have to completely disable the use of Cooperative Group in MGARD when compiling with HIP to make it work. So, I guess the support of Cooperative Group is just not mature enough at the current stage.

FYI:
Here is the bug report I submitted for zfp about Cooperative Group: https://github.com/RadeonOpenCompute/ROCm/issues/1729
Here is the bug report I submitted for MGARD also about Cooperative Group: https://github.com/RadeonOpenCompute/ROCm/issues/1645

CMakeLists.txt
JieyangChen7: Thanks. I will update this.

lindstro:
> There is no need to add explicit synchronization since we are calling synchronized memory copy API (internally it is implemented with fast asynchronized memory copy and ROCM runtime will add a synchronization for us)

My bad.  I think I mixed this up with some concerns I had with `cudaMallocAsync`, where "The allocation must not be accessed until the the allocation operation completes."  I guess there's no HIP equivalent.

src/hip_zfp/variable.h
```diff
@@ -192,12 +192,13 @@ namespace hipZFP
...
         {
             bool valid_stream = my_stream + i < nstreams_chunk;
-            unsigned long long offset0 = offsets[first_bitstream_block + i];
+            unsigned long long offset0 = 0;
             unsigned long long offset = 0;
             uint length_bits = 0;
             uint add_padding = 0;
             if (valid_stream)
             {
+                offset0 = offsets[first_bitstream_block + i];
```
lindstro, line 201: I've not yet fully absorbed how this code functions, but I can see how this might cause a memory violation for certain data/chunk sizes.  It seems clear that the same fix should apply to the CUDA implementation.  I will go ahead and merge this and apply the fix to CUDA myself.  Thanks for tracking down this bug!

lindstro, line 201: I've not yet fully absorbed how this code functions, but I can see how this might cause a memory violation for certain data/chunk sizes.  It seems clear that the same fix should apply to the CUDA implementation.  I will go ahead and merge this and apply the fix to CUDA myself.  Thanks for tracking down this bug!

lindstro:
LGTM

lindstro:
LGTM

PR #162, opened 2022-04-13
Fix exception handling in compress_numpy

jonashaag:
Without this a segfault would happen if there is an exception in `_init_field`


GarrettDMorrison:
LGTM

PR #167, opened 2022-08-02
Add __version__ attribute to zfpy module

cgohlke:


GarrettDMorrison:
We have plans in the near future to add zfpy version functionality akin to what the zfp version header provides which should provide what you're looking for here. Will try to get that in for you here once we have some free cycles.

PR #174, opened 2022-08-12
feat(python): add full header reader for zfp streams

william-silversmith:
Hello again!

I'm thinking of developing a compressed array reader for Python and I think this function would be helpful for identifying if the array is in fixed-rate mode. It would also be generally useful for inspecting zfp streams.

I hope you find it helpful!

Will


python/zfpy.pxd
```diff
@@ -31,7 +31,11 @@ cdef extern from "zfp.h":
         ptrdiff_t sx, sy, sz, sw
         void* data
     ctypedef struct zfp_stream:
-        pass
+        cython.uint minbits
```
lindstro, line 34: I would advise against exposing `zfp_stream`, whose definition here does not match the C definition and is certain to change in future versions.  A better solution is to query these parameters using `zfp_stream_params()`.

python/zfpy.pyx
```diff
@@ -351,3 +351,49 @@ cpdef np.ndarray decompress_numpy(
...
+
+    try:
+        if zfp_read_header(stream, field, HEADER_FULL) == 0:
+            raise ValueError("Failed to read required zfp header")
+
+        mode = zfp_stream_compression_mode(stream)
+        strm = stream[0]
+
+        return {
+            "nx": int(field.nx),
```
lindstro, line 379: Perhaps a dumb question from a Python novice: is there any concern here that this could overflow a Python 2 `int`, which is 32 bits?  Our metadata encoding of `zfp_field` allows 1D arrays with up to 2^48 elements. We may discuss whether or not to drop Python 2.

python/zfpy.pyx
```diff
@@ -351,3 +351,49 @@ cpdef np.ndarray decompress_numpy(
...
+            "ny": int(field.ny),
+            "nz": int(field.nz),
+            "nw": int(field.nw),
+            "type": ztype_to_dtype(field._type),
+            "sx": int(field.sx),
```
lindstro, line 384: These strides are not a property of the compressed data but rather of the layout of the NumPy array that the data is decompressed into.  I would suggest dropping these.

python/zfpy.pyx
```diff
@@ -351,3 +351,49 @@ cpdef np.ndarray decompress_numpy(
...
+            "sy": int(field.sy),
+            "sz": int(field.sz),
+            "sw": int(field.sw),
+            "mode": int(mode),
```
lindstro, line 388: Should we not return this as one of the enumerated types defined near the top?  Maybe we need a `zmode_to_dmode()` function analogous to `ztype_to_dtype()`?

python/zfpy.pyx
```diff
@@ -351,3 +351,49 @@ cpdef np.ndarray decompress_numpy(
...
+            "params": {
```
lindstro, line 389: Instead of only providing expert-mode parameters, I wonder if we could use something similar to the newly introduced `zfp_config` struct.

python/zfpy.pyx
william-silversmith: In Python2, Python `int` objects switch to C `long` from C `int` internally when they exceed their limit which should exceed the maximum size of the 1D array on 64-bit machines.

https://www.tutorialspoint.com/what-is-the-maximum-possible-value-of-an-integer-in-python

python/zfpy.pyx
william-silversmith: done

python/zfpy.pyx
william-silversmith: done

python/zfpy.pyx
william-silversmith: done

python/zfpy.pxd
william-silversmith: done

GarrettDMorrison:
@william-silversmith just wanted to check in on this with you. I think everything so far looks reasonable to if you get that conflict resolved we can get this PR merged.

william-silversmith:
I'm sorry, somehow I overlooked that that had happened. Fixed! Just waiting on testing to make sure it works.

lindstro:
LGTM.  @GarrettDMorrison Please approve and merge.

william-silversmith:
Just wanted to check in on this one. Once it's in, I'll be able to do more on making a Python compressed reader.

GarrettDMorrison:
@william-silversmith I've merged this in.

PR #197, opened 2023-01-05
Set HIP_PATH in CMakeLists.txt

thomasgibson:
This small PR simplifies the building of the HIP implementation of ZFP by setting the HIP_PATH automatically in the main CMakeLists.txt file, which avoids issues with generated Makefile having broken targets for `libamdhip64.so`. This avoids having to manually set `-DHIP_PATH=...` during configuration


lindstro:
LGTM

PR #223, opened 2024-01-10
Better argument typing

mirounga:
Consistent argument types for lift functions.

### DIFF
--- a/src/template/compress.c
+++ b/src/template/compress.c
@@ -18,7 +18,7 @@ _t2(compress, Scalar, 1)(zfp_stream* stream, const zfp_field* field)
 static void
 _t2(compress_strided, Scalar, 1)(zfp_stream* stream, const zfp_field* field)
 {
-  const Scalar* data = field->data;
+  const Scalar* data = (const Scalar*)field->data;
   size_t nx = field->nx;
   ptrdiff_t sx = field->sx ? field->sx : 1;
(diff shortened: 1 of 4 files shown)


lindstro:
@mirounga I agree with replacing `uint` with `ptrdiff_t`.  But what do the pointer casts accomplish?  Such casts are generally considered poor style in C.  Moreover, they may silently introduce bugs such as casting away const qualifiers that would be caught without the (unnecessary) cast.

mirounga:
Currently some functions have casts and some do not. In my experiments I have sometimes compiled the code as C++, and the absence of the cast had caused errors. If you consider casts bad idea, I could easily revert them.

lindstro:
For consistency and adherence to best practice, I would like to avoid any unnecessary casts involving `void*`.  If there are such casts elsewhere, can you please point them out?  I noticed that there are a few involving `malloc()` that were added over time for no obvious (to me) reason.

I really do not want to take on the burden of ensuring that the zfp C library can also be compiled with a C++ compiler without a compelling reason.  What is your reason for needing this, and are there not other changes that would be required?

Since this would be the first commit to `develop` following our last release, we'd also have to flag this as a development version.  Because the PR amounts to only two changed lines of code, it may be easiest if I simply take care of the necessary changes myself.  I do appreciate you pointing them out, however.

PR #226, opened 2024-02-19
Use explicit fallthroughs in encode.c

r-barnes:
Using
```
__attribute__((fallthrough));
```
allows the compiler to check for implicit fallthroughs with `-Wimplicit-fallthrough`, which has found _many_ issues in the codebases I've used it on.

It would be helpful if ZFP, as a core dependency, could use explicit fallthroughs to facilitate such checking.

### DIFF
--- a/src/template/encode.c
+++ b/src/template/encode.c
@@ -11,16 +11,16 @@ _t1(pad_block, Scalar)(Scalar* p, size_t n, ptrdiff_t s)
   switch (n) {
     case 0:
       p[0 * s] = 0;
-      /* FALLTHROUGH */
+      __attribute__((fallthrough));
     case 1:
       p[1 * s] = p[0 * s];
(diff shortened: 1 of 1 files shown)


lindstro:
Thanks for the suggestion.  This might be a good idea, but it would have to be done portably.  I believe this attribute (and `-Wimplicit-fallthrough`) is not supported prior to gcc 7, and presumably there are compilers other than MSVC that don't recognize it.  In C++17, it's `[[fallthrough]]`.  I would prefer some macro mechanism that can be applied to both C and C++ code.  Maybe add a `fallthrough_` macro to `include/zfp/internal/zfp/system.h`?

lindstro:
@r-barnes Take a look at 6814a4e and see if it meets your needs.

r-barnes:
@lindstro - That looks good, thanks!

Issue #178, opened 2022-08-27
`zfp` binary CUDA support broken?


tpwrules:
I can't figure out how to test CUDA support on my system.

I generated a simple test file: `python3 -c 'import struct; import math; open("test.bin", "wb").write(struct.pack("<256f", *list(math.sin(v/128*6.28)*1000 for v in range(256))))'`.

Then I tried to compress it:
`zfp -x cuda -i test.bin -z testz.bin -f -1 256 -r 10 -s -h`

Which reports the following info:
`type=float nx=256 ny=1 nz=1 nw=1 raw=1024 zfp=16 ratio=64 rate=0.5 rmse=707.3 nrmse=0.3536 maxe=1000 psnr=3.01`

Sometimes it reports nan values or random large floats, despite being given the same input data, but the statistics suggest it's not actually compressing and just storing zeros (related to #105 ?). If I remove the `-h` flag then it just tells me "compression failed". Everything works fine with the other execution modes (serial and omp).

I'm trying to build and package the 1.0.0 release, which doesn't come with tests, so I'm not sure if CUDA is broken in general or this is specific to the command line tool. I am building against CUDA 11.6.

lindstro:
I'm surprised that `zfp -x cuda -h` even works for you as the CUDA backend does not yet handle headers (`-h`) correctly (see #77).  Using your command line, I'm getting this message:
```
% zfp -x cuda -i test.bin -z testz.bin -f -1 256 -r 10 -s -h
incorrect or missing header
```
If you drop `-h`, you should be seeing this:
```
% zfp -x cuda -i test.bin -z testz.bin -f -1 256 -r 10 -s 
type=float nx=256 ny=1 nz=1 nw=1 raw=1024 zfp=320 ratio=3.2 rate=10 rmse=0.9085 nrmse=0.0004542 maxe=2.733 psnr=60.83
```
which matches what you get with the serial backend:
```
% zfp -i test.bin -z testz.bin -f -1 256 -r 10 -s
type=float nx=256 ny=1 nz=1 nw=1 raw=1024 zfp=320 ratio=3.2 rate=10 rmse=0.9085 nrmse=0.0004542 maxe=2.733 psnr=60.83
```
I'm assuming that you have built with `-DZFP_WITH_CUDA=ON`; you should be getting another error message if you haven't.

[...]

PR #232, opened 2024-06-27
update cuda cmake code

robertu94:
Hi @lindstro, some students I am working with encountered problems building the CUDA version of ZFP using a new version of CMake (v3.27 and newer).  This is because the FindCUDA cmake package was disabled in favor of FindCUDAToolkit and CUDA language support in this version and will be removed in a future version after being deprecated in 3.10.  The replacement features were introduced in cmake 3.17, but I find they are stable after 3.20 (the new cmake minimum required version that I introduced here).

As I have documented here: https://robertu94.github.io/guides/dependencies.html#tooling-versions, cmake 3.20 is available on all major Linux distros for their normal support cycles except Ubuntu 20.04, and CMake 3.20 is available on this platforms via `pip install cmake` or other package management systems including spack.  Supporting cmake 3.17 also includes CentOS-7 until it is deprecated in July, but will not support Ubuntu 20.04 which only provides cmake 3.16.

Regardless if you adopt this patch or not the spack package for ZFP needs to be updated as it does not build the cuda variant after cmake 3.27 to something of the effect of `depends("cmake@:3.26", when="@:1.0.0+cuda")`

cc: @jtian0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 wheelhouse
 zfpy.egg-info
 modules
+build-*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.20)
 
 # Enable MACOSX_RPATH by default
 cmake_policy(SET CMP0042 NEW)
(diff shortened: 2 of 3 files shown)


lindstro:
@robertu94 I appreciate the contribution and would like to merge your PR, though we first need to find a workaround for the AppVeyor tests, which use CMAKE 3.16.  I've been wanting to transition off AppVeyor and move to GitHub Actions.  I can't remember what the issues were when we last looked into this, and I'm afraid I don't have time to tackle that at the moment.

Another minor challenge is that we've been doing a lot of CUDA (and HIP) work on the staging branch (which has not been pushed to in a while).  This branch consolidates work from multiple branches, including new variable-rate (de)compression work and HIP support.  Once complete, we will merge all this new work into develop.

I would prefer to incorporate these changes there first and work out any potential kinks before merging into develop.  Currently our only GPU tests are being run at LLNL through GitLab CI, and so we'll also need to see if there are any issues there before I'm ready to merge into develop.

If you don't mind, can you please retarget your PR to the staging branch?  I expect it will be another couple of months or so before we will complete work on staging and will be ready to merge into develop.

lindstro:
@robertu94 A quick update: This PR conflicts with other changes we've recently made on the `staging` branch.  Some of your proposed changes have, however, been adopted in the latest commits on that branch, which also address #178.  This essentially required bumping CMake to 3.23, but only when `ZFP_WITH_CUDA` is enabled--otherwise we stick with CMake 3.9.  Please reopen if these changes do not address your needs.

robertu94:
Thanks @lindstro to close the loop, this fix works for me.

PR #234, opened 2024-07-02
Working pip install from setup.py

frobnitzem:
Explanation added to installation document.

Some setup.py options moved to pyproject.toml.
setup.py crashed when long_description
was left in though. So it was commented out.

Should consider setting license in pyproject.toml.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,9 @@ from setuptools import setup, Extension
 import numpy as np
 
 setup(
-    name="zfpy",
-    version="1.0.1",
-    author="Peter Lindstrom, Danielle Asher",
-    author_email="zfp@llnl.gov",
     url="https://zfp.llnl.gov",
-    description="zfp compression in Python",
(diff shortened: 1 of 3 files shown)


chiarua:
So glad to see what you've made.
Now I can make customized zfpy package depends on your modified setup.py.
Thank you.

setup.py
```diff
@@ -2,14 +2,9 @@
...
-    ext_modules=[Extension("zfpy", ["build/python/zfpy.c"],
+    #long_description="zfp is a compressed format for representing multidimensional floating-point and integer arrays. zfp provides compressed-array classes that support high throughput read and write random access to individual array elements. zfp also supports serial and parallel compression of whole arrays using both lossless and lossy compression with error tolerances. zfp is primarily written in C and C++ but also includes Python and Fortran bindings.",
+    ext_modules=[Extension("zfpy", ["python/zfpy.pyx"],
                            include_dirs=["include", np.get_include()],
-                           libraries=["zfp"], library_dirs=["build/lib64", "build/lib/Release"]), language_level = "3"]
```
lindstro, line 14: I believe older versions of Visual Studio use `build/lib/Release` as opposed to just `build/lib`.  For example, see AppVeyor output for a recent commit: https://open.cdash.org/viewBuildError.php?type=1&buildid=10034583.  Not sure if we need to preserve this.

PR #236, opened 2024-07-03
fix: syntax error in setup.py + modernize

william-silversmith:
Fixes the syntax error, cythonizes dynamically (may make this run on more platforms), sets language level to 3 (Python 2 is EOL for years now) to remove a warning, and adds license and classifiers so PyPI can categorize it better.

Loads numpy dynamically so it won't crash if numpy was installed at the same time as the setup call is run.

Uses numpy for installation, but oldest-supported-numpy for isolated compilation for py38 via pyproject.toml

Adds MANIFEST.in to ensure sdist includes source files, LICENSE, etc.

Resolves #233

PR #237, opened 2024-07-04
  fix: syntax error in setup.py + modernize

### LINKED ISSUE #233: Syntax error in setup.py
Syntax error (maybe misplaced parentheses)?

https://github.com/LLNL/zfp/blob/5c976d8da013988174f931845862b6f94119cade/setup.py#L14

```
% pip install --no-build-isolation 'git+https://github.com/LLNL/zfp.git@1.0.1'
Collecting git+https://github.com/LLNL/zfp.git@1.0.1
  Cloning https://github.com/LLNL/zfp.git (to revision 1.0.1) to /tmp/pip-req-build-aycbhn7n
  Running command git clone --filter=blob:none --quiet https://github.com/LLNL/zfp.git /tmp/pip-req-build-aycbhn7n
  Running command git checkout -q f40868a6a1c190c802e7d8b5987064f044bf7812
  Resolved https://github.com/LLNL/zfp.git to commit f40868a6a1c190c802e7d8b5987064f044bf7812
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [7 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-req-build-aycbhn7n/setup.py", line 14
          libraries=["zfp"], library_dirs=["build/lib64", "build/lib/Release"]), language_level = "3"]
                                                                                 ^^^^^^^^^^^^^^^^^^^^
      SyntaxError: invalid syntax. Maybe you meant '==' or ':=' instead of '='?
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

I'm using python 3.10.12 and numpy 2.0.0.

william-silversmith:
Fixes the syntax error, cythonizes dynamically (may make this run on more platforms), sets language level to 3 (Python 2 is EOL for years now) to remove a warning, and adds license and classifiers so PyPI can categorize it better.

Loads numpy dynamically so it won't crash if numpy was installed at the same time as the setup call is run.

Uses numpy for installation, but oldest-supported-numpy for isolated compilation for py38 via pyproject.toml

Adds MANIFEST.in to ensure sdist includes source files, LICENSE, etc.

Resolves https://github.com/LLNL/zfp/issues/233

This was moved from #236


da-wad:
Let me know when you merge this @lindstro ?

lindstro:
LGTM

lindstro:
@da-wad #237 has been merged.

setup.py
```diff
@@ -1,15 +1,48 @@
 from setuptools import setup, Extension
-import numpy as np
+
+class NumpyImport:
+  def __repr__(self):
+    import numpy as np
+
+    return np.get_include()
```
QuLogic, line 7: I don't understand the point of this class. I think it's meant to defer the import of `numpy`, but you call `str(NumpyImport())` in the arguments to `setup`, so it will be evaluated and import `numpy` at startup anyway. Was this change made redundant by adding `pyproject.toml` with build system requires?

pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
```
QuLogic, line 4: `wheel` is from some outdated documentation; you probably don't need it for your build. It should be automatically installed where necessary.

setup.py
lindstro: @william-silversmith Can you please address these questions?

william-silversmith: Yes, the point of the script is to delay the import of numpy. Originally, importing numpy concurrently with installation would cause the setup script to crash as numpy would not be found. This was a solution I found on StackOverflow. 

Later on, new python versions would not properly render the __repr__, so it was necessary to add str to get it to evaluate properly, but you might be right that it has rendered the delay mechanism superfluous.

lindstro: @william-silversmith Thanks for explaining (and for contributing this solution).

pyproject.toml
lindstro: @william-silversmith Here's a second question that you may be able to answer.  I unfortunately have minimal knowledge of Python, though I suspect it is related to our zfpy-wheels repo for building and uploading wheels to PyPI.

PR #245, opened 2024-11-16
Update cmake cuda code

robertu94:
Newer versions of cmake have deprecated the functions used in ZFP to find cuda.
Update to the new recommended way of doing things that is well supported on all
major distributions except for Ubuntu 20.04 out of the box.

- **update cuda cmake code**
- **add spack dev-build output to .gitignore**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 wheelhouse
 zfpy.egg-info
 modules
+build-*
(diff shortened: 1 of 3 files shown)


robertu94:
I would also appreciate it if you can release this change as 1.0.2

lindstro:
This appears to be a duplicate of PR #232.  See follow-ups there.

PR #255, opened 2025-02-28
Ignore ci warnings

vicentebolea:


lindstro:
@vicentebolea Thanks for your help with this.  I will merge once all tests pass.

FYI, it appears that the LLVM warning is caused by this:
https://github.com/LLNL/zfp/blob/f5999e0c241fff3c34afcc782b56035e34ab8d05/.github/workflows/tests.yml#L73-L80
I'm not entirely sure what the intent is here, but I believe this may have been an early workaround for dealing with poor OpenMP support with Apple clang.  Though the build matrix disables OpenMP with clang...

@GarrettDMorrison Do you know what's going on here?

Looking at the build output, `brew install libomp` outputs this:
```
For compilers to find libomp you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/libomp/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/libomp/include"
```
I think some tweaking is needed here; I'll look into it next week.

lindstro:
LGTM

PR #260, opened 2025-03-12
ci,cdash: rename cdash workflow to external

vicentebolea:


vicentebolea:
@lindstro the status check works!

vicentebolea:
It appears that it is posting the status as comming from your account. Make sure that it is using the robot token.

lindstro:
> It appears that it is posting the status as comming from your account. Make sure that it is using the robot token.

Hmm, that's odd.  I created another token from the robot account and updated.  Of course, the robot is not admin, so I had to update the token through my personal account.  Not sure if that has anything to do with it.

vicentebolea:
> Of course, the robot is not admin, so I had to update the token through my personal account. Not sure if that has anything to do with it

It should not be

vicentebolea:
@lindstro oh the new token is not associated to your account! I think that we completed this task! :)

vicentebolea:
Actually Ill go ahead and change the cdash job to another name so that it is not confusing.

vicentebolea:
@lindstro I reused this PR to rename the cdash workflow to external and the cdash job to generate_statuses. This will avoid the confusion of having two status checks named cdash.

PR #261, opened 2025-03-12
Testing CDASH status

vicentebolea:


vicentebolea:
@lindstro it worked! I think that we are finally done with this task :)

vicentebolea:
@lindstro it worked! I think that we are finally done with this task :)

lindstro:
@vicentebolea Fantastic!  Thanks for your help.  But should we expect only one job (ubuntu-20.04-g++-10) to show when following the CDash link?

vicentebolea:
The will all show up as in here, (master status) https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ZFP&showfilters=0&limit=100&value1=9f065ebe771b845fa212cc20bba20d0b95c180c2&showfeed=0

The thing is that I closed this too quickly, let me re-open to show it

PR #262, opened 2025-03-13
Fixes flaky cdash status

vicentebolea:
- Fixes the flaky cdash status by using the correct commit in both pull and pull_request_target events

### DIFF
--- a/.github/post-cdash-status
+++ b/.github/post-cdash-status
@@ -4,7 +4,7 @@ set -eo pipefail
 
 readonly API_BASE="https://api.github.com/repos/LLNL/zfp"
 readonly TOKEN="${CI_GITHUB_ROBOT_TOKEN}"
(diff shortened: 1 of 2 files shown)


lindstro (approved):
LGTM

vicentebolea:
@lindstro is there a way to cancel older/outdates appveyor builds?

lindstro:
> @lindstro is there a way to cancel older/outdates appveyor builds?

Yes, but it may be that only I can.  You have to be signed in to AppVeyor (e.g., through your GitHub account) to do this.  Should I cancel the build that's currently running (was started 3h ago)?

vicentebolea:
I think that I cannot do it since I am not part of LLNL org. Please cancel the previous ones

vicentebolea:
@lindstro since we are moving the win builds to gha I recycled this PR so that it only fixes de cdash status thing. I will merge and open a new one moving the wins builds to gha

PR #263, opened 2025-03-13
Move windows builds to github actions

vicentebolea:
This moves the windows build from Appveyor to Github.

IT now tests with:

- VS 16 2019 32b and 64b.
- VS 17 2022 32b and 64b
- Mingw 64b


vicentebolea:
@lindstro I can see that the appveyor jobs split the builds into two: One with OpenMP and one with cfp2. Are these opts mutually exclusive? Can we set them ON at the same time?

Thanks!

lindstro:
> @lindstro I can see that the appveyor jobs split the builds into two: One with OpenMP and one with cfp2. Are these opts mutually exclusive? Can we set them ON at the same time?

I believe we decided to split jobs up a very long time ago, perhaps because of Travis (?) time limits that we sometimes exceeded.  I can't think of any good reason to do so with our current test environment, however.

lindstro (approved):
LGTM

PR #265, opened 2025-03-20
ci: use cdash-status action

vicentebolea:
I have recently created a github action that setup the cdash status as shown here:https://github.com/marketplace/actions/cdash-status

In this PR remove the adhoc implementation of the cdash status and delegate into this action. This will help me not to duplicate effort across the different project that uses this.


lindstro (approved):
LGTM

lindstro:
@vicentebolea There's an [issue](https://github.com/LLNL/zfp/actions/runs/13997891172) with the workflow.  Can you please investigate?

vicentebolea:
fixed here #266

PR #266, opened 2025-03-21
ci: fix version in cdash-status

vicentebolea:
@lindstro sorry for the noise, we can only test this cdash-status ci jobs in master. This should be the last change.


lindstro (approved):
LGTM

PR #267, opened 2025-04-07
ci: correct cdash-status-url

vicentebolea:
We have recently moved the cdash-status action to the Kitware namespace. This PR updates its URL.


lindstro (approved):
LGTM

PR #271, opened 2025-06-15
Fix zfp_stream_maximum_size overflow on 32bit target

### LINKED ISSUE #270: zfp_stream_maximum_size can overflow on wasm32
I've compiled ZFP to wasm32 and experienced out-of-bounds writes. In my particular case, I'm compressing a 9900 x 16987 2D array of single precision floats with an absolute error tolerance of 0.0402.

Some back of the napkin math based on https://github.com/LLNL/zfp/blob/7d1e3a21047a976599b562b3bbd53b1f34348f1a/src/zfp.c#L740-L770

brings me to:

dims = 2

bx = 2475
by = 4247
blocks = 10511325

values = 16

maxbits = 0
maxbits += 9;
maxbits += 16 - 1 + 16 * 32;
maxbits = 536

148 + 10511325 * 536 + stream_word_bits - 1
      ^^^^^^^^^^^^^^
Here is where the multiplication overflows the size_t (32bit on wasm32), even though the division by CHAR_BIT would bring the result back into a valid 32bit integer.

Since the correct zfp_stream_maximum_size result would fit into a 32bit integer, i.e. the request is not excessive, I think the zfp_stream_maximum_size function should correctly handle the overflow here.

CC @milankl and @treigerm

juntyr:
Fixes #270

### DIFF
--- a/src/zfp.c
+++ b/src/zfp.c
@@ -766,7 +766,12 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
   maxbits += values - 1 + values * MIN(zfp->maxprec, zfp_field_precision(field));
   maxbits = MIN(maxbits, zfp->maxbits);
   maxbits = MAX(maxbits, zfp->minbits);
-  return ((ZFP_HEADER_MAX_BITS + blocks * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
+  
+  uint64 maxsize = ((ZFP_HEADER_MAX_BITS + ((uint64)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
+
+  if (maxsize > SIZE_MAX)
+    return 0;
+  return (size_t)maxsize;
 }
 
 void
--- a/src/zfp.c
+++ b/src/zfp.c
@@ -769,7 +769,8 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
   
   bitstream_size maxsize = ((ZFP_HEADER_MAX_BITS + ((bitstream_size)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
 
-  if (maxsize > SIZE_MAX)
+  /* check if the maxsize fits into size_t to avoid silent truncation */
+  if (((size_t)maxsize) != maxsize)
     return 0;
   return (size_t)maxsize;
 }
--- a/src/zfp.c
+++ b/src/zfp.c
@@ -767,7 +767,7 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
   maxbits = MIN(maxbits, zfp->maxbits);
   maxbits = MAX(maxbits, zfp->minbits);
   
-  uint64 maxsize = ((ZFP_HEADER_MAX_BITS + ((uint64)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
+  bitstream_size maxsize = ((ZFP_HEADER_MAX_BITS + ((bitstream_size)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
 
   if (maxsize > SIZE_MAX)
     return 0;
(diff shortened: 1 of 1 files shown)


src/zfp.c
```diff
@@ -766,7 +766,12 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
   maxbits += values - 1 + values * MIN(zfp->maxprec, zfp_field_precision(field));
   maxbits = MIN(maxbits, zfp->maxbits);
   maxbits = MAX(maxbits, zfp->minbits);
-  return ((ZFP_HEADER_MAX_BITS + blocks * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
+  
+  uint64 maxsize = ((ZFP_HEADER_MAX_BITS + ((uint64)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
```
lindstro: `uint64` here ought to be `bitstream_offset`

src/zfp.c
```diff
@@ -766,7 +766,12 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
...
+
+  if (maxsize > SIZE_MAX)
```
lindstro: `SIZE_MAX` was introduced in C99.  We'll need a more general fix for C89.  Let me think about it.

src/zfp.c
juntyr: `bitstream_size` might be even better since we're computing a size here (and it's just an alias to `bitstream_offset`), right?

src/zfp.c
juntyr: Could we use `(size_t)(-1)` instead?

lindstro: Yes, but the cleanest way would probably be to add
```
#ifndef SIZE_MAX
  #define SIZE_MAX ((size_t)-1)
#endif
```
at the end of [this](https://github.com/LLNL/zfp/blob/7d1e3a21047a976599b562b3bbd53b1f34348f1a/include/zfp/internal/zfp/types.h#L58) section.

Some compilers emit warnings when casting a signed literal to unsigned, so there may be variants like `((size_t)0 - 1)` that are needed to suppress such warnings.  I believe we have similar code elsewhere in zfp.  For now, the above may suffice to correct the bug.

juntyr: Though https://stackoverflow.com/a/44401992 suggests that some pre-processors don't support this? I don't know enough about C to know it that's correct so I'll go with your judgement

lindstro: What it means is that you can't write preprocessor code like
```
#if SIZE_MAX == 0xfffffffffffffffful
```
because the `(size_t)` cast isn't allowed in such preprocessor expressions.  Of course, if `SIZE_MAX` wasn't defined in the first place, that code would also fail (or give an unexpected result; `SIZE_MAX` would expand to 0).  So you shouldn't include `zfp.h` to define `SIZE_MAX` for you and then rely on using it in preprocessor directives.

This said, I'm a little wary of introducing such code that works only some of the time.  Possibly the right approach is to deal with data models in a more comprehensive way, which we've already begun some work on.  But it will be some time before we roll that code out.

There are possibly other ways of dealing with this, such as
```
if (maxsize != (size_t)maxsize)
```
which would be true if `maxsize` does not fit in a `size_t`.

juntyr: I like that, let's use that!

juntyr:
@lindstro Is there anything else that needs to be done on this PR?

src/zfp.c
```diff
@@ -766,7 +766,13 @@ zfp_stream_maximum_size(const zfp_stream* zfp, const zfp_field* field)
   maxbits += values - 1 + values * MIN(zfp->maxprec, zfp_field_precision(field));
   maxbits = MIN(maxbits, zfp->maxbits);
   maxbits = MAX(maxbits, zfp->minbits);
-  return ((ZFP_HEADER_MAX_BITS + blocks * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
+  
+  bitstream_size maxsize = ((ZFP_HEADER_MAX_BITS + ((bitstream_size)blocks) * maxbits + stream_word_bits - 1) & ~(stream_word_bits - 1)) / CHAR_BIT;
```
lindstro: Can't mix code and variable declarations in C89.

On a separate note, we ought to have a test that compiles zfp using a C89 compiler.

lindstro:
LGTM

PR #273, opened 2025-08-07
Enable CMake languages in ZFP

nischild:
I encountered an error when adding ZFP to [this project](https://gitlab.mpcdf.mpg.de/bsl6d/bsl6d/-/blob/119-data-compressed-halo-exchange/cmake/BSL6DZfpConfig.cmake?ref_type=heads) using FetchContent

If ZFP is build with CUDA capabilities this leads to a delayed error stating missing CMAKE_CUDA_CREATE_STATIC_LIBRARY, which seems to be a configuration option for windows.
```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CUDA_CREATE_STATIC_LIBRARY
```

This can be avoided if CUDA is enabled as a language explicitly in the project. This error is thrown only, if `enable_language(CUDA)` is called at a later stage which in our case is done by the HeFFTe subproject.

Minimal reproducable with cmake/4.0, gcc/12, cuda/12.2:
```sh
cmake -S . -B build -D CMAKE_CXX_COMPILER=g++ -D CMAKE_C_COMPILER=gcc
```
[minimalZFP_fetchContent.zip](https://github.com/user-attachments/files/21666279/minimalZFP_fetchContent.zip)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ else()
     "${ZFP_VERSION_MAJOR}.${ZFP_VERSION_MINOR}.${ZFP_VERSION_PATCH}.${ZFP_VERSION_TWEAK}")
 endif()
 
-project(ZFP VERSION ${ZFP_VERSION})
+project(ZFP VERSION ${ZFP_VERSION} LANGUAGES C CXX)
 
 #------------------------------------------------------------------------------#
 # Some boilerplate to setup nice output directories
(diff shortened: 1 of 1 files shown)


lindstro:
We've made a bunch of changes on the `staging` branch with respect to CUDA.  [One of those changes](https://github.com/LLNL/zfp/blob/daf60368eae42c21c0a4783fe004ddf3aa8da8a5/CMakeLists.txt#L227-L232) is the addition of `enable_language(CUDA)` that hopefully will address the issue you're seeing.  We hope to merge this branch into `develop` in the near future and then release new CUDA features this fall.

Note that the `staging` branch is currently very much a work in progress, and I would advise against relying on it until it has been merged into `develop`.

nischild:
Yes, that should solve it

PR #276, opened 2025-09-03
fix typo ?

jeis4wpi:
does this fix make sense regarding the variable typo?

Noticed some trailing whitespace.


lindstro:
Yes, not sure how we missed this.  Thanks for fixing.

lindstro:
LGTM

PR #283, opened 2025-12-12
Pin GHA dependencies by hash

williamjallen:
Hash-pinned dependencies are an important requirement enforced by the OpenSSF scorecard tool: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

This commit pins all GitHub Actions workflows to a commit hash.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -11,10 +11,10 @@ jobs:
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
 
       - name: Setup Python
-        uses: actions/setup-python@v6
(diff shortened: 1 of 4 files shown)


lindstro:
Thanks for your contribution.  If I understand correctly, adding such hashes would preclude automatic PRs when new versions of actions like `checkout` and `python-setup` are rolled out.  It probably makes sense to uses hashes for "less trusted" workflow dependencies.

williamjallen:
@lindstro Dependabot can update hash-pinned dependencies, [including the comment](https://github.com/dependabot/dependabot-core/pull/5951).  Given that you already have Dependabot configured to run weekly, there shouldn't be any negative effects on your workflow.  An extra bonus of pinning everything is that you can add GHA check to verify that everything is pinned, preventing unpinned dependencies from making it into your workflows in the future.  That said, if you don't want to pin the `actions/*` dependencies, I'd be happy to revert those changes and pin only the 3rd-party actions.

lindstro:
LGTM

PR #287, opened 2026-01-25
Restrict workflow permissions to minimum required

williamjallen:
Limiting the scope of the GITHUB_TOKEN via the [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions) attribute is best practice to prevent a leaked GITHUB_TOKEN from allowing repository write access.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -5,6 +5,9 @@ on:
     workflows: [Tests]
     types: [completed]
 
+permissions:
+  contents: read
+
 jobs:
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/debug-linux.yml
+++ b/.github/workflows/debug-linux.yml
@@ -2,6 +2,9 @@ name: Debug (Linux)
 
 on: [workflow_dispatch]
 
+permissions:
+  contents: read
+
 jobs:
     debug:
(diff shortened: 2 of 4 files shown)


lindstro:
@williamjallen Thanks for your help.  This change makes sense to me.

lindstro:
LGTM

lindstro:
@williamjallen After merging this PR, our LLNL GitLab CI tests are no longer being reflected in the test status, even though they are being run.  Can you think of any reason for this?  For comparison, see the test status for commit a888ab7606c58cc5ab5f34298fec25473e0d28ad, which shows that `ci/gitlab/lc.llnl.gov` passed.

williamjallen:
Hi @lindstro, that certainly seems like an odd coincidence.  I can't find anything that seems even remotely relevant unfortunately.  Can you provide some more details about how that GitLab status check gets sent back?

You can share details with me privately at `william.allen@kitware.com` if necessary.

lindstro:
After further investigation, this appears to be an issue with the GitLab/GitHub integration on our end when posting the test status.  Sorry for the false alarm.

lindstro:
Just as an FYI, this issue was caused by a change to LLNL security policy.  We got it all sorted out and our GitLab CI pipelines are now reporting back to GitHub.

